PR #13, opened 2020-06-11
v0.10.0

vincentaudoire:


src/lib/scss/private/base/spacing.stories.mdx
```diff
@@ -0,0 +1,37 @@
+<Meta title="Base/Spacing" />
+
+# Spacing
+
+Dirty swan provides way to express spacing
```
killthekitten, line 5: Could you explicitly tell about the 8px increment rule here?

vincentaudoire: I don't think it should be explicitly told here, it's a design system wide requirement and not especially around the margin-top. 

A more ideal place would be to have it set at the intro section. Because if I start to add it here I'll need to add this for with `width and height`, `sizes` and every time there will be something affected by the multiple of 8 rule.

Also this is less relevant here as you just need to know what are the different spacing capabilities.

killthekitten: 👍

PR #17, opened 2020-10-29
[FEA-681] Create MultiDropzone component for multiple doc uploads

kseha1080:


src/lib/components/multiDropzone/UploadFileCell/index.tsx
```diff
@@ -0,0 +1,134 @@
+import React from 'react';
+import classnames from 'classnames';
+
+import styles from './style.module.scss';
+import icons from '../icons/index';
+import { UploadStatus, UploadedFile } from '..';
+
+const getUploadingIcon = (type: string) => {
```
vincentaudoire, line 8: would be cleaner if you define `type IconType = 'heic' | 'bmp'…` and then the function parameter would be `type: string | IconType`.

Then you can ensure type safety and proper mapping if you add a new icon

src/lib/components/multiDropzone/index.tsx
```diff
@@ -0,0 +1,88 @@
...
+        })}
+        {...getRootProps()}
+      >
+        <input {...getInputProps()} />
+        <img src={icons.uploadIcon} alt="" />
+        <div className="p-h4 mt8">
+          {disabled
+            ? 'Please wait while uploading file...'
+            : 'Choose file or drag & drop'}
+        </div>
```
vincentaudoire, line 68: `disabled` doesn't seem to be a good name here, as far as I can see when disabled the text "Please wait while uploading file..." is being displayed. Maybe a better wording would be `uploading`?

src/lib/components/multiDropzone/style.module.scss
```diff
@@ -0,0 +1,29 @@
...
+.container {
+  background-color: white;
+  padding: 24px;
+  box-shadow: 0px 2px 3px rgba(38, 38, 46, 0.04);
+  border-radius: 8px;
+
+  @include p-size-mobile {
+    padding: 16px;
+  }
+}
```
vincentaudoire, line 13: I think the outline must be set somewhere so that it fits our guideline.

image

Look at the outline set on a button for a good example

vincentaudoire: btw: not sure the outline needs to be set on the `container`

vincentaudoire (changes requested):

vincentaudoire:
There's no visual confirmation when hovering with a selected file - I think you might want to sync up with Matt and find a way to do it. 

drop

PR #20, opened 2020-11-02
[NoTicket] Export models

kseha1080:


vincentaudoire (approved):

src/lib/models/downloadRing/index.ts
```diff
@@ -0,0 +1,6 @@
+export type DownloadStatus =
```
vincentaudoire, line 1: I think you might want to the status to the component itself. Maybe `DownloadRingDownloadStatus` would be more fitted?

vincentaudoire (changes requested):

src/lib/models/downloadRing/index.ts
kseha1080: How about `DownloadRingStatus`?

PR #24, opened 2020-12-07
DateSelector component bug fix

kseha1080:
I reconfigured setting states to more specific points on input.
Updates were tested on other screens as well so existing screens should be fine.
The React compiler warnings while rendering and setting state on component are fixed as well.

### DIFF
--- a/src/lib/components/dateSelector/index.tsx
+++ b/src/lib/components/dateSelector/index.tsx
@@ -1,12 +1,12 @@
-import React, { useState } from "react";
-import moment from "moment";
-import { CalendarDate } from "@popsure/public-models";
+import React, { useState, useEffect } from 'react';
+import moment from 'moment';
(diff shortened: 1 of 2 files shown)


src/lib/components/dateSelector/index.tsx
```diff
@@ -63,41 +63,49 @@ const DateSelector = ({
   const availableMonths = moment.monthsShort();
 
   const [date, setDate] = useState<Partial<CalendarDate>>(
-    CalendarDateValue ?? {}
+    CalendarDateValue ?? {},
```
vincentaudoire, line 66: It look like it shouldn't be starting with a capital letter (these are only for react components). I know you're not the one that made this change but can you update it so it's not so confusing?

vincentaudoire (changes requested):

PR #47, opened 2021-04-20
Cha 1811 multi select suggest component

kseha1080:


src/lib/components/chipList/chip/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+        onClick={onRemove}
+      >
+        <img
+          src={
+            hoveredRemoveButton === true
+              ? removeButtonHighlightedIcon
+              : removeButtonIcon
+          }
+          alt="removal x button"
+        />
```
vincentaudoire, line 40: You can achieve this with `mask-image` in pure css.

src/lib/components/chipList/chip/index.tsx
```diff
@@ -0,0 +1,44 @@
+import { useState } from 'react';
+
+import styles from './style.module.scss';
+import removeButtonIcon from '../icons/remove-button.svg';
+import removeButtonHighlightedIcon from '../icons/remove-button-highlighted.svg';
+import { Option } from '../../../models/suggestMultiSelect';
+
+export default ({
```
vincentaudoire, line 8:
```suggestion
const Chip = ({
```

Maybe the chip is a component of its own that can be used as-is?

src/lib/components/chipList/index.stories.mdx
```diff
@@ -0,0 +1,29 @@
...
+  <ChipList
+    values={[
+      { value: 'Private Health' },
+      { value: 'Public Health' },
+      { value: 'Dental' },
+      { value: 'Legal' },
+    ]}
+    onRemove={() => {}}
+  />
+</Preview>
```
vincentaudoire, line 29: Could you show an example with the `img` option being set?

src/lib/models/suggestMultiSelect/index.ts
```diff
@@ -0,0 +1,4 @@
+export interface Option {
+  value: string;
+  img?: string;
```
vincentaudoire, line 3:
```suggestion
  leftIcon?: string;
```

Would `leftIcon` be a better name?

vincentaudoire (changes requested):

vincentaudoire (approved):

PR #49, opened 2021-05-04
[NoTicket] Update colors

kseha1080:


vincentaudoire (approved):
Stating the obvious here but since it's a big change:
This has been checked and triple checked by Andrey?

Also that would be a major version bump (as it's going to change quite a few elements)

PR #50, opened 2021-05-05
Bugfix on bottom or regular modal

vincentaudoire:


src/lib/components/modal/bottomOrRegularModal/index.tsx
```diff
@@ -1,18 +1,42 @@
...
+  }, []);
+
   return (
     <>
-      <div className={styles.mobile}>
-        <BottomModal {...props} />
+      <div className={styles.mobile} ref={mobileRef}>
+        {visibleSize === 'mobile' && <BottomModal {...props} isOpen={isOpen} />}
       </div>
       <div className={styles.desktop}>
```
kseha1080, line 36: Wouldn't this be an empty `div` if `visibleSize` is mobile?

kseha1080: Perhaps `{visibleSize === 'desktop' && <div className={styles.desktop}><RegularModal /></div>}` is better?

vincentaudoire: Yes good point

kseha1080 (approved):

PR #61, opened 2021-06-21
Cha 2204 extend card component

kseha1080:
This PR is for a new variation of the `CardWithTopIcon` component.
There's a breaking change where the description of the component is now updated to solely React `children` components without any styles - this gives more freedom to style it at a higher level.

[See component in Figma](https://www.figma.com/file/582xiOgwHGIl4y5kGHrQ69/App?node-id=2277%3A0)

### DIFF
--- a/src/lib/components/cards/cardWithTopIcon/index.tsx
+++ b/src/lib/components/cards/cardWithTopIcon/index.tsx
@@ -32,8 +32,8 @@ export default ({
       {...props}
     >
       <img
-        width={topIconSize?.width}
-        height={topIconSize?.height}
+        width={topIconSize.width}
+        height={topIconSize.height}
         alt={topIcon.alt}
         src={topIcon.src}
       />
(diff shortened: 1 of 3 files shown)


src/lib/components/cards/cardWithTopIcon/index.tsx
```diff
@@ -29,7 +31,12 @@ export default ({
       } ${className ?? ''}`}
       {...props}
     >
-      <img width="48xp" height="48px" alt={topIcon.alt} src={topIcon.src} />
+      <img
+        width={topIconSize ? `${topIconSize?.width}px` : '48xp'}
```
vincentaudoire, line 35: It'll be better if you set `topIconSize` with a default value of `48px`, then it'll save you this ternary operator.

vincentaudoire (changes requested):

vincentaudoire (approved):

PR #65, opened 2021-06-22
upgrading dependencies

vincentaudoire:


src/lib/components/autocompleteAddress/index.tsx
```diff
@@ -175,7 +175,7 @@ const AutoCompleteAddress = ({
             service.getDetails(
               { placeId: firstResult.place_id },
               (newPlace) => {
-                onPlaceChanged(newPlace, false);
+                onPlaceChanged(newPlace ? newPlace : undefined, false);
```
kseha1080, line 178: I think instead of ternary, TS operator `??` can be used?

kseha1080 (changes requested):

src/lib/components/autocompleteAddress/index.tsx
vincentaudoire: good point 👍

kseha1080 (approved):

PR #68, opened 2021-07-16
[CHA-2532] add required prop to allow closing the

talo242:
This PR adds the following functionality:

- Adds a new `required` prop to prevent the user from closing the modal
- Adds a new `touchmove` event handler to prevent scroll on mobile devices
- Moves all the component logic to the `useOnOpen` hook to share it with the different modals

### DIFF
--- a/src/lib/components/modal/index.stories.mdx
+++ b/src/lib/components/modal/index.stories.mdx
@@ -26,6 +26,7 @@ Modals are dialog overlays that prevent the user from interacting with the rest 
 | onClose   | `function () => void` | Callback when the user close the modal                                                 | n/a           | true     |
 | children  | React.jsx             | The content that gets displayed on the modal                                           | n/a           | true     |
 | className | string                | Any additional className                                                               | n/a           | false    |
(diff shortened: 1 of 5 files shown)


src/lib/components/modal/index.stories.mdx
```diff
@@ -26,6 +26,7 @@ Modals are dialog overlays that prevent the user from interacting with the rest
 | onClose   | `function () => void` | Callback when the user close the modal                                                 | n/a           | true     |
 | children  | React.jsx             | The content that gets displayed on the modal                                           | n/a           | true     |
 | className | string                | Any additional className                                                               | n/a           | false    |
+| required  | boolean               | Prevent the user from closing the modal                                                | false         | false    |
```
vincentaudoire, line 29: I'm not sure about the name, might be a bit confusing… what do you think about `dismissable`?

talo242: Yeah, sounds better. But it would be inversed right?
The default value should be `true` and if set to `false` the user will not be able to close the modal.

vincentaudoire: Indeed!

PR #73, opened 2021-08-09
Add flex utilities and responsiveness v0.1

talo242:
This PR adds utilities for the most common display properties and a very basic example:

It also introduces the following spacing utilities:
- `p{x}` for adding padding utilities.
- `mx{x}`, `my{x}`, `p{x}` and `py{x}` for adding left-right and bottom-top spacing.

Lastly, it adds a mixin for working on responsive styles, and also adds the first version of it to the `display` and `spacing` utilities.

The proposed API is using the prefixes `md:` and `lg:` to add styles to medium and large devices respectively.
I.E. `<div className="p16 md:p24 lg:p32">` will apply the following classes to the `div` component:

```css
.p16 {
  padding: 16px;
}

@media (min-width: 45rem) and (max-width: 64rem) {
  .md\:p24 {
    padding: 24px;
  }
}

@media (min-width: 64rem) {
  .lg\:p32 {
    padding: 32px;
  }
}
```
There is a working example on the Flex page.

One thing to consider: right now using the `@include grid.add-responsive` will introduce a new media query declaration for each CSS property. I'm worried this could make our generated CSS file grow very very big.
I think the best approach would be adding a compression that groups all media query declarations. I would like to know your opinion on this matter.

### DIFF
--- a/src/lib/scss/private/base/_flex.scss
+++ b/src/lib/scss/private/base/_flex.scss
@@ -1,0 +1,43 @@
+@use 'grid';
+
+/** Shared values across align and justify properties */
+$common-values: (
+  'start': flex-start,
+  'end': flex-end,
+  'center': center,
+);
+
+$align-values: map-merge(
+  $common-values,
+  (
+    'stretch': stretch,
+    'baseline': baseline,
+  )
+);
+
+$jutify-values: map-merge(
+  $common-values,
+  (
+    'between': space-between,
+    'around': space-around,
+    'evenly': space-evenly,
+  )
+);
+
+// Align items
+@each $name, $value in $align-values {
+  .ai-#{$name} {
+    @include grid.add-responsive {
+      align-items: $value;
+    }
+  }
+}
+
+// Justify content
+@each $name, $value in $jutify-values {
+  .jc-#{$name} {
+    @include grid.add-responsive {
+      justify-content: $value;
+    }
+  }
+}
--- a/src/lib/scss/private/base/_grid.scss
+++ b/src/lib/scss/private/base/_grid.scss
@@ -1,4 +1,5 @@
 @use '../../public/grid';
+@use '../utils/str-replace';
 
 .p-body {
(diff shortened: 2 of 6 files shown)


vincentaudoire:
I like the approach very much! 
I'm not sure I fully understand what you want to do by "adding a compression that groups all media query declarations", could you elaborate a bit on that?
Additionally we can output the generated bundle size with all the css files as part of a step when creating the pull request and see the difference in kb between the two. But I agree that this needs to be used widely as it can grow out of proportion very quickly.

talo242:
> I'm not sure I fully understand what you want to do by "adding a compression that groups all media query declarations", could you elaborate a bit on that?

What I mean is instead of having multiple `@media` declarations for each new class:
```css
@media (min-width: 45rem) and (max-width: 64rem) {
  .md\:ai-start {
    align-items: flex-start;
  }
}

@media (min-width: 64rem) {
  .lg\:ai-start {
    align-items: flex-start;
  }
}

@media (min-width: 45rem) and (max-width: 64rem) {
  .md\:ai-end {
    align-items: flex-end;
  }
}

@media (min-width: 64rem) {
  .lg\:ai-end {
    align-items: flex-end;
  }
}
```
We could have all classes with the same media queries grouped together:
```css
@media (min-width: 45rem) and (max-width: 64rem) {
  .md\:ai-start {
    align-items: flex-start;
  }

  .md\:ai-end {
    align-items: flex-end;
  }
}

@media (min-width: 64rem) {
  .lg\:ai-start {
    align-items: flex-start;
  }
  
  .lg\:ai-end {
    align-items: flex-end;
  }
}

```
I did a little research and it seems that SASS doesn't tackle this issue, so we have to find another solution.

> Additionally we can output the generated bundle size with all the css files as part of a step when creating the pull request and see the difference in kb between the two. But I agree that this needs to be used widely as it can grow out of proportion very quickly.
This is a very good idea, I'll look into it 👍🏼

PR #74, opened 2021-08-12
[CHA-2706] Add Card Button component

kseha1080:
**Updated:** Ellipsis is removed for flexibility

### DIFF
--- a/src/lib/components/cards/cardButton/index.tsx
+++ b/src/lib/components/cards/cardButton/index.tsx
@@ -1,15 +1,13 @@
-import { FormEvent, createElement } from 'react';
+import { FormEvent, createElement, ReactNode } from 'react';
 
 import styles from './style.module.scss';
 import { arrowRight } from '../icons';
-import classnames from 'classnames';
 
 interface Props {
   title: string;
-  description: string;
+  description: string | ReactNode;
   disabled?: boolean;
   className?: string;
-  ellipsis: boolean;
 }
 
 type ActionProps =
@@ -19,22 +17,14 @@ type ActionProps =
 const CardContent = ({
   title,
   description,
-  ellipsis,
 }: {
   title: string;
-  description: string;
-  ellipsis: boolean;
+  description: string | ReactNode;
 }) => (
   <>
-    <div className="w90">
+    <div>
       <div className="p-p--small">{title}</div>
-      <div
-        className={classnames('tc-primary-500 p-p', {
-          [styles['text-ellipsis']]: ellipsis,
-        })}
-      >
-        {description}
-      </div>
+      <div className="tc-primary-500 p-p">{description}</div>
     </div>
     <img {...arrowRight} />
   </>
@@ -47,7 +37,6 @@ export default ({
   onClick,
   href,
   className,
-  ellipsis = false,
 }: Props & ActionProps) => {
   const component = href ? 'a' : 'button';
   return (
@@ -57,13 +46,7 @@ export default ({
           className ?? ''
         }
         `,
-        children: (
-          <CardContent
-            title={title}
-            description={description}
-            ellipsis={ellipsis}
-          />
-        ),
+        children: <CardContent title={title} description={description} />,
         disabled,
         ...(component === 'a' ? { href } : { onClick }),
       })}
--- a/src/lib/components/cards/cardButton/index.tsx
+++ b/src/lib/components/cards/cardButton/index.tsx
@@ -2,14 +2,14 @@ import { FormEvent, createElement } from 'react';
 
 import styles from './style.module.scss';
 import { arrowRight } from '../icons';
-import { limitTextLength } from '../../../util/limitTextLength';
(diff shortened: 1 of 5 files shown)


src/lib/util/limitTextLength.test.ts
```diff
@@ -0,0 +1,7 @@
+import { limitTextLength } from './limitTextLength';
+
+describe('Limit text length testing', () => {
```
vincentaudoire, line 3: Would you be able to use `text-overflow: ellipsis;` instead?

kseha1080: That was my first implementation but `text-overflow: ellipsis` limits text to one line. There will be cases in the future where descriptions will need to partially displayed but more than one line. Do you see any potential issues with implementing it this way?

vincentaudoire: I'm not sure but I believe `text-overflow: ellipsis` has some accessibility features behind it. @talo242 wdyt?

src/lib/components/cards/cardButton/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  onClick,
+  href,
+  maxChar,
+  className,
+}: Props & ActionProps) => {
+  const component = href ? 'a' : 'button';
+
+  return (
+    <>
+      {createElement(component, {
```
talo242, line 52: 👌🏼

src/lib/util/limitTextLength.test.ts
talo242: Yeah, it's more accessible because all the text is in the HTML and it's only being visually hidden using CSS.
The bad thing with ellipsis is that we cannot specify how many characters we want to display, we have to rely on the container's width.

I think we can still use the current approach and add a `title` attribute to the element. That way screen readers can read the whole text and also if someone hovers over it it will display the whole text.

src/lib/components/cards/cardButton/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+  maxChar,
+}: {
+  title: string;
+  description: string;
+  maxChar?: number;
+}) => (
+  <>
+    <div>
+      <div className="p-p--small">{title}</div>
+      <div className="tc-primary-500 p-p">
```
talo242, line 31:
```suggestion
      <div className="tc-primary-500 p-p" title={description}>
```

src/lib/components/cards/cardButton/index.tsx
```diff
@@ -19,17 +19,21 @@ type ActionProps =
...
+    <div className="w90">
       <div className="p-p--small">{title}</div>
-      <div className="tc-primary-500 p-p">
-        {maxChar ? limitTextLength(description, maxChar) : description}
+      <div
+        className={classnames('tc-primary-500 p-p', {
+          [styles['text-ellipsis']]: ellipsis,
+        })}
+      >
+        {description}
```
vincentaudoire: What about doing another approach here? Instead of having `ellipsis: boolean;` you can accept description as either `string` or `React.ReactNode`. If it's of style string you render it with `p-p tc-primary-500` and if it's of type node you just render the component. It's a bit more flexible I would say as you live the ellipsis implementation to the parent component.

Wdyt?

kseha1080: I'm not sure I understand what you're trying to achieve, why would we need a JSX type of `description` prop if the parent will be implemented with ellipsis anyway? Could you give an example?

src/lib/components/cards/cardButton/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  <>
+    <div>
+      <div className="p-p--small">{title}</div>
+      <div className="tc-primary-500 p-p">{description}</div>
+    </div>
+    <img {...arrowRight} />
+  </>
+);
+
+export default ({
```
talo242, line 33: Let's just name this default export to prevent eslint from complaining.

src/lib/components/cards/cardButton/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  title,
+  description,
+}: {
+  title: string;
+  description: string | ReactNode;
+}) => (
+  <>
+    <div>
+      <div className="p-p--small">{title}</div>
+      <div className="tc-primary-500 p-p">{description}</div>
```
talo242, line 27: I think what @vincentaudoire meant was adding `className="tc-primary-500 p-p"` only when the description prop is a `string`. If the description is a `ReactNode` we should render the prop without the `div` wrapper.

talo242 (changes requested):

src/lib/components/cards/cardButton/index.tsx
vincentaudoire: Yes, exactly 👍

vincentaudoire (approved):

PR #75, opened 2021-08-16
[CHA-2590] Create card with checkbox component

talo242:
This PR adds a new Card with checkbox component using the current Card and Checkbox components:
Screen Shot 2021-08-16 at 14 15 56


vincentaudoire (approved):
I'm going to challenge the fact that this is going to be widely used enough that it needs to be part of dirty-swan?

kseha1080 (approved):

PR #78, opened 2021-08-19
[CHA-2833] Create the SegmentedControl component

vincentaudoire:
This is the implementation of the Segmented Control component 👉 [Figma reference file](https://www.figma.com/file/MKs4cbojdVOBKUxv7okb93/Dirty-Swan-Design-System?node-id=234%3A13)

There's the possibility for a single lined segmented control and for a double lined segmented control. It's possible to navigate the component with the keyboard (using TAB and ENTER)

https://user-images.githubusercontent.com/3629231/130113286-1010c044-f875-499e-8bf1-202f102bd25e.mov

### DIFF
--- a/src/lib/components/segmentedControl/index.tsx
+++ b/src/lib/components/segmentedControl/index.tsx
@@ -1,0 +1,102 @@
+import { useEffect, useRef, useState } from 'react';
+import styles from './style.module.scss';
+
+interface TitleWithSubtitle {
+  title: string;
+  subtitle: string;
+}
+
+const SegmentedControl = ({
+  className = '',
+  values,
+  selectedIndex,
+  onChange,
+}: {
+  className?: string;
+  values: Array<TitleWithSubtitle> | Array<string>;
+  selectedIndex: number;
+  onChange: (selectedIndex: number) => void;
+}) => {
+  const [selectedChipBackgroundWidthLeft, setSelectedChipBackgroundWidthLeft] =
+    useState<{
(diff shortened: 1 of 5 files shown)


src/lib/components/segmentedControl/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+  values: Array<TitleWithSubtitle> | Array<string>;
+  selectedIndex: number;
+  onChange: (selectedIndex: number) => void;
+}) => {
+  const [selectedChipBackgroundWidthLeft, setSelectedChipBackgroundWidthLeft] =
+    useState<{
+      left: number;
+      width: number;
+    }>({ left: 0, width: 0 });
+  const chipContainer = useRef<any | null>(null);
```
talo242, line 25: I don't know how strict we want to be with TS, but here we could use HTMLDivElement type to avoid using `any`.
That would mean an extra verification in the `useEffect` hook, but could potentially prevent type errors if this code is edited in the future.

talo242 (approved):

src/lib/components/segmentedControl/index.tsx
vincentaudoire: You're right, I've missed that part (I casted to any and do the proper typing later but I just forgot 😓 )
We could add a strict rule on the linter to avoid that.

kseha1080 (approved):

talo242 (approved):

PR #80, opened 2021-08-25
Cha 2683 extend input error styles

kseha1080:


src/lib/components/input/index.stories.mdx
```diff
@@ -37,5 +52,14 @@ You are looking at the JSX definition of the Input component, if you want you ca
     <Input className="wmx5 mt8" placeholder="amount" prefix="€" />
     <h4 className="p-h4 mt24">Filled with placeholder</h4>
     <Input className="wmx5 mt8" placeholder="amount" value="100" prefix="€" />
+    <h4 className="p-h4 mt24">Error</h4>
+    <Input
+      className="wmx5 mt8"
+      placeholder="amount"
+      value="100"
+      prefix="€"
+      hasError
```
talo242, line 61: What do you think of having only an `error` prop that receives a string?
```suggestion
      error="Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi ac hendrerit justo. Nunc nibh ex, vulputate ac leo id, feugiat rhoncus purus."
```
I cannot think of a situation where only one of those props is used, I think it will reinforce best practices to always provide an error message to show the error state.

PR #93, opened 2021-11-08
[STO-3198] Add grey-250 and update input disabled color

kseha1080:


vincentaudoire (commented):
do we really need a 250? 😅

talo242:
I agree with Vincent. I would suggest that we stick to 300 to show disabled inputs because 250 is very close to the 200 value, so it could not be clear enough that the input is disabled (In my humble opinion, I think 200 is way too light and not too accessible).

Also, if we really need to add these "middle values", maybe we can use SASS modifiers instead of creating lots of new colors.

vincentaudoire:
> > do we really need a 250? 😅
> 
> It is quite a collection of colors, but for disabled inputs, 300 is a bit too dark and 200 is too light for contrast against font color. We can also use this in the future for other disabled fields as well.

I would seriously push back on this one - it's too easy to add new colours. Maybe we should tweak the 200 or the 300 then? 

Because tomorrow we may have to add a 225 or 275 which is defeating the whole point of having this design system.

PR #98, opened 2022-02-04
Kerem/emu 3827 move download button to dirty swan from

keremciu:
This PR includes several changes:

### Adds

- DownloadButton component - this is already used in app project inside Feather, there's requirement to use it in a different project that's why it's moved here.

### Fixes

- Several console errors fixed

Screenshot 2022-02-04 at 17 23 26


src/lib/components/downloadButton/index.tsx
```diff
@@ -0,0 +1,68 @@
+// import { email } from '@getpopsure/private-constants';
```
talo242, line 1: I think we can remove this comment

talo242 (approved):

PR #112, opened 2022-03-04
fix(npm): [STO-3642] Add src folder to package.json's tracked files

kimkwanka:
Currently only the dist folder is included in the package which makes react-scripts 5.0 show 51 warnings after the planned version bump in the app repo.

image

Adding the src folder to package.json's "files" should get rid of this issue:

https://github.com/facebook/create-react-app/discussions/11767#discussioncomment-2096794

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   "bin": "dist/bin/index.js",
   "files": [
     "dist",
+    "src",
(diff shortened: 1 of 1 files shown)


kseha1080 (approved):
👍

PR #113, opened 2022-03-04
feat(git): [STO-4025] Add PR template to DS repo

kimkwanka:
### What this PR does

This PR adds the PR template of app and website to the DS repo

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,30 @@
+### What this PR does
+
+Please include a summary of the change and which issue is fixed or a link to the feature it implements.
+
+### Why is this needed?
+
+Please include relevant motivation and context of why this PR is necessary, sentry/linear/notion/...
+
+Solves:  
+STO-XXX  
+EMU-XXX
+
+### Checklist:
+
+- [ ] I reviewed my own code
+- [ ] The changes align with the designs I received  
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,30 @@
...
+
+Please include a summary of the change and which issue is fixed or a link to the feature it implements.
+
+### Why is this needed?
+
+Please include relevant motivation and context of why this PR is necessary, sentry/linear/notion/...
+
+Solves:  
+STO-XXX  
+EMU-XXX
```
talo242, line 11: Because this is an open-source project, I think we can skip this part.
What do you think?

kimkwanka: I don't actually think having it in there is really detrimental as most (if not all) PRs are from Feather people. This means in most cases it would be useful and might only be unnecessary in a minuscule amount of scenarios. And even then, people can just delete those lines, same as they'd do with all the other placeholder / description text in the template.

Maybe we could add "OPTIONAL" or something similar instead just to make it clear that it's not mandatory to have this in the PR?

talo242: Yeah, you are right. Let's keep it as it is 😄

talo242 (approved):

kseha1080 (approved):

PR #119, opened 2022-04-27
fix(bundle): Now using public-model 1.0.0 and esm module

vincentaudoire:
### What this PR does

1. We're updating to `public-model` in order to take advantage of the ESM module definition
2. We're exporting this library as ESM (as opposed to CJS) module to optimize the consumer's bundle size

### Why is this needed?

- The `dist/index.js` file size was 3MB 😱 now it is 682KB 
 - NextJS12 now supports ESM, it's now safe to have all our libraries exported as ESM.

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Edge

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "Readme.md"
   ],
   "dependencies": {
-    "@popsure/public-models": "^0.0.14",
+    "@popsure/public-models": "^1.0.0",
     "classnames": "^2.3.1",
     "dayjs": "^1.10.7",
     "lodash.debounce": "^4.0.8",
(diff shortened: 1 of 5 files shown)


keremciu:
It is important to upgrade jest version 28 to make ESM work in tests

react-scripts version 5 includes jest 28

vincentaudoire:
> It is important to upgrade jest version 28 to make ESM work in tests
> 
> react-scripts version 5 includes jest 28

Actually, it looks like jest 28 doesn't quite yet support ESM https://jestjs.io/blog/2022/04/25/jest-28

Maybe the way to go would be to both provide CJS and ESM as part of `public-model` to maintain backward compatibiity?

PR #120, opened 2022-04-27
chore: Updating @popsure/public-model to 1.0.1

vincentaudoire:
# What this PR does

We're updating to public-model in order to take advantage of the ESM module definition

# Why is this needed?

- The dist/index.js file size was 3MB 😱 now it is 733KB
- NextJS12 now supports ESM, it's now safe to have all our libraries exported as ESM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@popsure/dirty-swan",
-  "version": "0.26.13",
+  "version": "0.27.0-alpha.3",
   "author": "Vincent Audoire <vincent@getpopsure.com>",
   "license": "MIT",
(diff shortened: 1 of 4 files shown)


vincentaudoire:
@keremciu I redid this one with public model 1.0.1 (still in alpha). The major change is in public model 1.0.1 where we are exporting both CJS modules and ESM modules. Therefore the test fallback on the CJS module and we don't need to touch anything 🎉

kimkwanka (approved):

talo242 (approved):

kimkwanka (approved):

PR #128, opened 2022-07-08
feat(localization): [STO-4281] Expose props to enable localization

kimkwanka:
### What this PR does

1.  Fixes the uneven widths for the `AutocompleteAddress` on mobile
**BEFORE:**
image

**AFTER:**
image

2. Exposes placeholder text and other props to make `AutocompleteAddress` and `DateSelector` components localizable

image
image

3. Updates the docs

### Why is this needed?

Solves:  
STO-4281  
STO-3809  

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Edge

### DIFF
--- a/src/lib/components/dateSelector/index.tsx
+++ b/src/lib/components/dateSelector/index.tsx
@@ -1,7 +1,6 @@
 import { useState, useEffect } from 'react';
 import dayjs from 'dayjs';
 import 'dayjs/locale/de';
-import 'dayjs/locale/bs';
 
 import localeData from 'dayjs/plugin/localeData';
 import { CalendarDate } from '@popsure/public-models';
@@ -14,6 +13,8 @@ import {
 import styles from './style.module.scss';
 import './datepicker.scss';
 import calendarIcon from './icons/calendar.svg';
+
+export type supportedDayJSLocales = 'en' | 'de';
 
 dayjs.extend(localeData);
 const COLLECTABLE_DATE_FORMAT = 'YYYY-MM-DD';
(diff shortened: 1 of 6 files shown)


src/lib/components/dateSelector/index.tsx
```diff
@@ -12,8 +14,9 @@ import styles from './style.module.scss';
 import './datepicker.scss';
 import calendarIcon from './icons/calendar.svg';
 
-dayjs.extend(localeData);
+export type supportedDayJSLocales = 'en' | 'de';
```
ruizhen88, line 17: Not important, just adding my opinion - IMO this name can be more concise by saving `supported`, because I feel like `supported` is already implied by enum, limiting it to a certain set of values.

ruizhen88 (approved):

src/lib/components/dateSelector/index.tsx
kimkwanka: To be honest, I wanted it to be a bit more verbose so that in the future, when someone tries using this component inside the other repos and stumbles across this property it provides more context. `dayJSLocales` for example wouldn't be able to convey the whole story as it is not an enum of all the dayjs locales but only a subset of those, the ones that are supported by the component (because they were imported). Even if you'd see that there's only 2 entries currently in the union type, you could come to the conclusion that the type itself might just not be up to date and try to pass in another of dayjs' locales.
But yeah with names there's probably never a 100% correct answer and you can find good arguments for either side really. But in my experience it rarely hurts to be a bit "too verbose".

ruizhen88: > But in my experience it rarely hurts to be a bit "too verbose".
Yes, agreed!

PR #141, opened 2022-07-26
[STO-4342] Refactor Multidropzone to collect multiple files at once

kseha1080:
### What this PR does

1. Refactors Multidropzone component to collect multiple files at once
2. Adds `maxFiles` prop to the component to give control of how many files can be uploaded
3. Adds a generic error display to the component when user tried to upload more files than allowed.

‼️ Note: This component should be refactored again to unify all errors. Currently, file size and invalid file type errors are handled on BE and therefore omitted from defining the error inside this component.

### Why is this needed?

We want to allow customers to upload multiple files at once

Solves:  
STO-4342

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Edge

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "hastscript": "^7.0.2",
     "lodash.debounce": "^4.0.8",
     "lodash.isequal": "^4.5.0",
+    "react-animate-height": "^3.0.4",
     "react-autosuggest": "^10.1.0",
     "react-day-picker": "^7.4.10",
     "react-dropzone": "^11.3.2",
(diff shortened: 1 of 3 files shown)


talo242:
@kseha1080 do you think it's necessary to allow this via configuration? i.e. passing a prop?
I'm wondering wether this change will trigger bugs wherever only one file is being expected.

kseha1080:
> @kseha1080 do you think it's necessary to allow this via configuration? i.e. passing a prop? I'm wondering whether this change will trigger bugs where only one file is being expected.

I checked for this situation and couldn't find any, but it would be a nice addition to the component. I'll update the component to allow control over this.

talo242 (approved):

kimkwanka (approved):

talo242 (approved):
👍🏼

ruizhen88 (approved):

PR #143, opened 2022-08-05
fix(markdown): Add missing export

mbaguiar:
### What this PR does

Adds another missing export 🙈

### Checklist:

- [x] I reviewed my own code
  Or give a reason why this does not apply: No visual changes
  Or give a reason why this does not apply: No visual changes

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -29,9 +29,14 @@ export {
   TableRowHeader,
   TableInfoButton,
   SegmentedControl,
-  DownloadButton
+  DownloadButton,
+  Markdown,
 } from './lib';
 
-export type { TableHeader, DownloadRingDownloadStatus, DownloadStatus } from './lib';
+export type {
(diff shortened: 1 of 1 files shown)


talo242 (approved):

talo242:
I think it's not super clear the places you have to include new features to export them correctly. Maybe we could look into having just one export file or updating the docs to add clearer instructions 🤔 
What do you think @mbaguiar, what was your experience right now adding this new feature?

mbaguiar:
> I think it's not super clear the places you have to include new features to export them correctly. Maybe we could look into having just one export file or updating the docs to add clearer instructions 🤔 What do you think @mbaguiar, what was your experience right now adding this new feature?

Yeah, I definitely feel we should include this somewhere. And do we really need the two index files? It was pretty frustrating to figure out I didn't add the export two times 😅

PR #146, opened 2022-09-28
chore: [no-ticket] remove stale components

ruizhen88:
### What this PR does

Remove components that are not used in applications:
Dropzone, DownloadRing, CardWithTopIcon

### Checklist:

- [x] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge


kseha1080 (approved):
This task has been chilling in backlog forever. Thanks for taking this up ✌️

src/index.tsx
```diff
@@ -3,20 +3,17 @@ import App from './App';
...
   MultiDropzone,
-  DownloadRing,
   IbanInput,
   CurrencyInput,
   BottomModal,
   RegularModal,
   BottomOrRegularModal,
   CardWithTopLeftIcon,
   CardWithLeftIcon,
-  CardWithTopIcon,
```
kimkwanka, line 19: Good stuff! ✨ The only one I would be careful with, would is probably `CardWithTopIcon`. Could you double check with the designers that this component is not used in the designs? Even if the component is not used in our codebases it doesn't necessarily mean it should be removed, it might just mean that people cooked up their own version of a component that already exists. In that case, refactoring the code to actually use it would be the more appropriate call.

kimkwanka: It definitely looks like something that might have been used in the designs:
image

kimkwanka: There was some discussion about cards in general in terms of design (which lead to [this ticket](https://linear.app/feather-insurance/issue/STO-3862/split-card-ds-component-in-two) being closed) so it would be great to have the designers' opinion on this matter.

kimkwanka (approved):

src/index.tsx
ruizhen88: Ayse did mention this design is being used, but when I examined the use case, it didn't seem like it could directly use this DS component. But I see your point, I will leave it here and we could update it to make it more reusable, then refactor the cards in website and app that fall into this category.

PR #148, opened 2022-10-26
feat(ComparisonTable): [NoTicket] Update implementation

kimkwanka:
### What this PR does

This PR merges the current implementation of the `ComparisonTable` with the very similar [CoverageModal](https://github.com/getPopsure/app/blob/8d155d02f5c13bb0fd8a3a59187fd7ee4162a1b3/src/components/CoverageDetails/CoverageModal/CoverageModal.tsx) component used in app which will allow us to replace both eventually. As this is designed as a drop-in-replacement for the ComparisonTable component first and foremost however, the `CoverageModal` will need changing of data structure etc. to be able to utilize this new component.

image

More detailed list of features of this PR:

1. Fix incorrect behaviour (incorrectly aligned elements, non-functional arrow buttons, etc.) in some contexts where reliance on viewport units for calculations fails
2. Add collapsible sections
3. Allow first column and content columns / cells to have sizes other than the previous defaults to better match the new designs. In conjunction with the growContent option this allows for more graceful transitions from desktop to mobile sizes.
4. Add the ability to add a y-offset to the sticky header to be able to use it on website where there's already a sticky nav bar

### Why is this needed?

To fix existing errors in the ComparisonTable while being able to replace the CoverageModal tables with it later down the line.

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Edge


ruizhen88 (approved):

src/lib/components/comparisonTable/components/AccordionItem/AccordionItem.module.scss
```diff
@@ -0,0 +1,44 @@
+@import '../../../../scss/public/colors';
+
+.container {
+  width: var(--growContent, min-content);
+  margin: 0 auto;
+}
+
+.container + .container {
```
kseha1080, line 8: Is this a typo?

src/lib/components/comparisonTable/style.module.scss
```diff
@@ -12,6 +12,20 @@
...
 
+.noScrollBars {
+  // Disable the scrollbar in all browsers
+  scrollbar-width: none; /* Firefox */
+  -ms-overflow-style: none; /* Internet Explorer 10+ */
+  &::-webkit-scrollbar {
+    /* WebKit */
+    width: 0;
+    height: 0;
+  }
```
kseha1080, line 23: Don't we need horizontal scroll bars?

kimkwanka: It's limited to the table content, it does not disable scroll bars globally.

src/lib/components/comparisonTable/components/AccordionItem/AccordionItem.module.scss
kimkwanka: No, it's meant to add distance between containers ("if a container follows a container")

kimkwanka: Which means the very first container won't have that margin.

kseha1080 (approved):

PR #159, opened 2022-12-27
fix(SegmentedControl): [NoTicket] add key property

busypeoples:
### What this PR does

* Add missing unique key property
* Add simple test to verify click and key down functionality

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge


ruizhen88 (approved):

talo242 (approved):
Awesome! Thanks for this update.

I think DS dependencies need an update (like using `user-events` instead of `fireEvent`) but this is something that we can talk about during Frontend Jour Fixe.

PR #162, opened 2023-01-19
 fix(input): [NoTicket] add stable id to input

busypeoples:
### What this PR does

Make the generated input id stable and use a simple id generation approach, as the id only has to be unique at render time.

* Drop the uuid dependency 

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/input/index.tsx
+++ b/src/lib/components/input/index.tsx
@@ -1,8 +1,11 @@
-import React from 'react';
+import React, { useState } from 'react';
 import classnames from 'classnames';
-import { v4 as uuidv4 } from 'uuid';
 
 import styles from './style.module.scss';
+
+const generateUniqueId = () => {
+  return `input-id-${Math.floor(Math.random() * 10000000)}`;
+};
 
 // Something weird is going on with enterKeyHint that makes it a required field under certain circumstances. The & Omit<…> and & Pick<…> is a hacky way to go around that.
 export type InputProps = {
@@ -29,7 +32,7 @@ export default React.forwardRef(
     }: InputProps,
     ref?: React.ForwardedRef<HTMLInputElement>
   ) => {
-    const uniqueId = id ?? uuidv4();
+    const [uniqueId] = useState(id ?? generateUniqueId());
     return (
       <div className={`${styles.container} ${className ?? ''}`}>
         {label && (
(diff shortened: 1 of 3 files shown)


src/lib/components/input/index.tsx
```diff
@@ -29,7 +32,7 @@ export default React.forwardRef(
     }: InputProps,
     ref?: React.ForwardedRef<HTMLInputElement>
   ) => {
-    const uniqueId = id ?? uuidv4();
+    const [uniqueId] = useState(id ?? generateUniqueId());
```
talo242, line 35: Why do we need a state here? Is only having a computed value not enough?

kimkwanka: Is there any harm in dropping that id generation altogether? As far as I can tell the id is forwarded to the underlying `input` if an id would be needed and an `htmlFor` prop could be `similarly` exposed instead.

kimkwanka: If it is an a11y concern then I would argue that a manually set `id` is a lot more helpful to screen readers etc, than a randomly generated string.

busypeoples: @kimkwanka we could go with the implicit label route, but remember reading that using ids is better than implicit, but I might be mistaken.

kimkwanka: https://dev.to/thejaredwilcurt/the-truth-behind-implicitexplicit-form-labels-2e10

Seems like they should in theory be equal, but apparently there are some screen readers that do not work nicely with implicit labels. Guess we can go with random labels then :)

kimkwanka: Still not sure I like the potential overlap with Math.random here but then again I am not sure if alternatives like [nanoid](https://github.com/ai/nanoid) have the same stability issues as uuid

kimkwanka (approved):

src/lib/components/input/index.tsx
busypeoples: Agreed @kimkwanka 
Maybe we can do it in two steps.
Merge this to fix the uuid dependency issue and then we could try to go the implicit label route or discuss it.

PR #164, opened 2023-01-23
refactor(input): [no-ticket] update input to use sr-only class on hidden label

diogomateus:
### What this PR does
Update input to use sr-only class on hidden label. 
This PR also adds the `.sr-only` class to dirty-swan utils.

The `sr-only` class is based on several utils and accessibility recommendations:
https://tailwindcss.com/docs/screen-readers
https://a11y-guidelines.orange.com/en/articles/accessible-hiding/

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/input/currency/index.stories.mdx
+++ b/src/lib/components/input/currency/index.stories.mdx
@@ -27,5 +27,7 @@ The following will be done:
     <CurrencyInput className="wmx5 mt8" label="Amount" placeholder="100" />
     <h4 className="p-h4 mt24">With label and no placeholder</h4>
     <CurrencyInput className="wmx5 mt8" label="Amount" />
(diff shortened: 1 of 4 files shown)


ruizhen88 (approved):
🙌

busypeoples (approved):
This looks like a reasonable approach to define a specific screen reader class.
Would also be interesting to hear what @kimkwanka and @kseha1080 think about it as well.

kimkwanka:
As outlined in the [Slack thread](https://feather-insurance.slack.com/archives/C031Y4Q0A22/p1674489013745109) I feel like this is something that can be helpful in the future to alleviate a11y issues in somewhat tricky scenarios. 👍

kimkwanka (approved):

PR #165, opened 2023-01-25
fix(tree-shaking): [NoTicket] enable tree-shaking

busypeoples:
### What this PR does

This is an initial approach to enable consuming apps to tree-shake ds correctly.

To help tree-shake optimized apps, we will need to make some changes:
* `package.json` needs to define an esm specific entry point, this can be achieved by defining `module`.
* `package.json` needs a defined `sideEffects` which excludes css to ensure that the the library can be tree-shaked.
* Create an `esm` build that is isolated from the `cjs` build.
* Set the `TypeScript` output directory to the `esm` build folder to enable consuming applications to find the types.

These are first steps to get to a minimal bundle size when working with ds.

#### Testing the changes

To `test` the changes, the following approach is required:

* Create a controlled environment, f.e. using vite and the react-ts template to setup a barebones app
* Run `yarn add ../dirty-swan` (assuming the test app and ds are siblings, otherwise the relative path to ds)
* Add one or two ds components
* Run in dev mode and see if it compiles and check the console output
* Run the build and check the bundled js size
* Use an analyzer library that can help with visualizing the bundle (i.e. `rollup-plugin-visualizer`)

#### Testing results

Testing the changes yields the following results when testing it with a barebones `vite:react-ts` setup:

The example App imports four ds components:

```
import {
  Chip,
  Input,
  CurrencyInput,
  Button,
} from "@popsure/dirty-swan";
```

#### Pre changes

tree-shaking-pre-1

tree-shaking-pre-2

#### Post changes

tree-shaking-post-1

tree-shaking-post-2

tree-shaking-post-3

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -20,28 +20,29 @@ const postcssCustomProperties = require('postcss-custom-properties');
  * Need to look into which plugins are needed for each build.
  *
  */
+const plugins = [
+  external(),
+  postcss({
+    modules: true,
+    plugins: [postcssCustomProperties({ preserve: true })],
+  }),
+  url(),
+  resolve(),
+  commonjs(),
+];
+
 export default [
   {
     input: 'src/lib/index.tsx',
     output: [
       {
-        dir: 'dist',
+        dir: 'dist/cjs',
         format: 'cjs',
         exports: 'named',
         sourcemap: true,
       },
     ],
-    plugins: [
-      external(),
-      postcss({
-        modules: true,
-        plugins: [postcssCustomProperties({ preserve: true })],
-      }),
-      url(),
-      resolve(),
-      typescript(),
-      commonjs(),
-    ],
+    plugins: [...plugins, typescript({ outDir: 'dist/cjs' })],
   },
   {
     input: Object.fromEntries(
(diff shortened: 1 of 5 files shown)


vincentaudoire (approved):

rollup.config.js
```diff
@@ -4,27 +4,66 @@ import external from 'rollup-plugin-peer-deps-external';
...
+        exports: 'named',
+        sourcemap: true,
+      },
+    ],
+    plugins: [...plugins, typescript({ outDir: 'dist/cjs' })],
+  },
+  {
+    input: Object.fromEntries(
+      glob
+        // we need to look into why adding ts breaks the build
```
kimkwanka, line 50: Is it maybe picking up test files when adding ".ts"?

kimkwanka (approved):
✨

rollup.config.js
busypeoples: @kimkwanka  This could actually be the case but also it seems we don't need it at the moment.

PR #168, opened 2023-01-27
refactor(multidropzone): [EMU-3320] Add accept filetype validation support

diogomateus:
### What this PR does
Add accept filetype validation support to MultiDropZone component.

1. Adds support to accept prop on MultiDropZone component, including helper states of `document` and `image`
2. Refactors component to have a separated file for utils and types
3. Supports showing file errors both inline and as a general state
4. Adds textOverrides support to error messages from react-dropzone

Solves:  
EMU-3320

https://user-images.githubusercontent.com/4015038/215064138-3886a175-eaf8-408b-8a6b-3d4faabbd12f.mov

### Checklist:

- [x] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,8 @@
     "react-scroll-sync": "^0.9.0",
     "remark-directive": "^2.0.1",
     "sass": "^1.35.1",
-    "signature_pad": "^3.0.0-beta.3"
+    "signature_pad": "^3.0.0-beta.3",
+    "uuid": "^9.0.0"
   },
   "peerDependencies": {
     "react": "^16.12.0",
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "remark-directive": "^2.0.1",
     "sass": "^1.35.1",
     "signature_pad": "^3.0.0-beta.3",
-    "uuid": "^9.0.0"
+    "uuid": "^8.3.2"
   },
   "peerDependencies": {
     "react": "^16.12.0",
--- a/src/lib/components/multiDropzone/index.tsx
+++ b/src/lib/components/multiDropzone/index.tsx
@@ -40,7 +40,7 @@ const MultiDropZone = ({
   uploading,
   onRemoveFile,
   isCondensed = false,
-  maxFiles = 2,
+  maxFiles = 0,
   textOverrides,
 }: Props) => {
   const [errors, setErrors] = useState<ErrorMessage[]>([]);
--- a/src/lib/components/multiDropzone/index.tsx
+++ b/src/lib/components/multiDropzone/index.tsx
@@ -5,57 +5,34 @@ import AnimateHeight from 'react-animate-height';
 import styles from './style.module.scss';
 import icons from './icons/index'; // TODO: inline all of the svgs
 import UploadFileCell from './UploadFileCell';
-
-export type UploadStatus = 'UPLOADING' | 'COMPLETE' | 'ERROR';
-
-export type FileType =
-  | 'heic'
-  | 'bmp'
-  | 'jpeg'
-  | 'jpg'
-  | 'png'
-  | 'doc'
-  | 'docx'
-  | 'pdf';
-
-const getUploadStatus = (progress: number, error?: string): UploadStatus => {
-  if (error) {
-    return 'ERROR';
-  }
-
-  if (progress < 100) {
-    return 'UPLOADING';
-  }
-
-  return 'COMPLETE';
-};
-
-export interface UploadedFile {
-  id: string;
-  name: string;
-  type: FileType | string;
-  previewUrl?: string;
-  progress: number;
-  error?: string;
-  showProgressBar?: boolean;
-  showLoadingSpinner?: boolean;
-}
+import { 
+  formatAcceptFileList, 
+  getErrorMessage, 
+  getFormattedAcceptObject, 
+  getUploadStatus 
+} from './utils';
+import { 
+  AcceptType, 
+  ErrorMessage, 
+  FileType, 
+  TextOverrides, 
+  UploadedFile, 
+  UploadStatus 
+} from './types';
 
 interface Props {
+  accept?: AcceptType;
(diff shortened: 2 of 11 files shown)


src/lib/components/multiDropzone/types.ts
```diff
@@ -0,0 +1,38 @@
+import { Accept } from "react-dropzone";
+
+export type UploadStatus = 'UPLOADING' | 'COMPLETE' | 'ERROR';
+
+export type FileType =
```
kimkwanka, line 5: If you define this union type based on const string arrays like this:
```
export const DOCUMENT_FILES = ['doc', 'docx', 'pdf'] as const;
export const IMAGE_FILES = ['heic', 'bmp', 'jpeg', 'jpg', 'png'] as const;

export const FILE_TYPES = [...DOCUMENT_FILES, ...IMAGE_FILES] as const;
export type FileType = typeof FILE_TYPES[number];
```
you can then reuse those arrays later in your functions like `getFormattedAcceptObject`. This way you can be sure that your exported type matches the accepted file extensions. 
This means that all you need to do to add more supported files in the future, is add them to either `IMAGE_FILES` or `DOCUMENT_FILES` and everything will work out of the box with no further changes required.

kimkwanka (approved):

src/lib/components/multiDropzone/types.ts
diogomateus: Sadly `getFormattedAcceptObject`/`accept` needs to have extensions prefixed with a `.` so I can't reuse it 😢

src/lib/components/multiDropzone/index.tsx
```diff
@@ -64,23 +41,35 @@ export default ({
...
 
   const onDrop = useCallback(
     (acceptedFiles: File[], filesRejected: FileRejection[]) => {
-      setError('');
+      resetError();
 
       if (filesRejected.length > 0) {
-        setError(filesRejected[0].errors[0].message);
+        setError(getErrorMessage(
+          filesRejected[0].errors[0],
```
talo242, line 56: What will happen if I drag and drop multiple files and more than one of them are not valid?
If I see correctly, this implementation will only take the error of the first rejected file.
Do you think it'd be a better idea to have an array of errors?

src/lib/components/multiDropzone/types.ts
talo242: I like @kimkwanka's idea, I don't see a problem in attaching the`.` to the filename in order to be correctly formatted.
You could also use the same array on the `formatAcceptFileList`.

What do you think?

talo242 (changes requested):
I think we also need to start thinking about how to test this component.
Maybe we can write some unit tests for the utils and an integration test for more complex testing.
What do you think?

src/lib/components/multiDropzone/types.ts
kimkwanka: Just add a dot :D

diogomateus: I'll update, I was removing the dots on another place anyway, I can add on the other cases instead 😄

package.json
```diff
@@ -34,7 +34,8 @@
     "react-scroll-sync": "^0.9.0",
     "remark-directive": "^2.0.1",
     "sass": "^1.35.1",
-    "signature_pad": "^3.0.0-beta.3"
+    "signature_pad": "^3.0.0-beta.3",
+    "uuid": "^9.0.0"
```
busypeoples, line 38: There were some issues with `uuid`, downgrading might help. Best is to test it with the website and see if the tests go through.

diogomateus: Thanks for the heads up, just downgraded the version, also for consistency among projects.

src/lib/components/multiDropzone/utils/index.test.ts
```diff
@@ -0,0 +1,118 @@
...
+});
+
+describe('getFormattedAcceptObject', () => {
+  it('Should return image accept object if is of type image', () => {
+    expect(getFormattedAcceptObject("image")).toEqual({
+      'image/heic': [".heic"],
+      'image/bmp': [".bmp"],
+      'image/jpeg': [".jpeg"],
+      'image/jpg': [".jpg"],
+      'image/png': [".png"]
```
talo242, line 44: Can you use the `imagesAccept` variable here to keep clarity?

src/lib/components/multiDropzone/types.ts
```diff
@@ -0,0 +1,34 @@
+import { Accept } from "react-dropzone";
+
+export type UploadStatus = 'UPLOADING' | 'COMPLETE' | 'ERROR';
+
+export const DOCUMENT_FILES = ['doc', 'docx', 'pdf'];
+export const IMAGE_FILES = ['heic', 'bmp', 'jpeg', 'jpg', 'png'];
+
+export const FILE_TYPES = [...DOCUMENT_FILES, ...IMAGE_FILES];
+export type FileType = typeof FILE_TYPES[number];
```
talo242, line 9: ✨

src/lib/components/multiDropzone/types.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export interface UploadedFile {
+  id: string;
+  name: string;
+  type: FileType | string;
```
talo242, line 14: Is there a reason to leave `string` here? I feel like doing so removes the safety from the interface.
What do you think?

talo242: Fair point. I would keep it strict and add more types as needed, but maybe we can asses this in another PR just to be able to merge this one.

talo242 (approved):

src/lib/components/multiDropzone/types.ts
diogomateus: I'll create an issue to revisit this, we can probably add some proper way of checking types and mime types

PR #169, opened 2023-01-27
refactor(multidropzone): [EMU-5671] Add file maxSize validation support

diogomateus:
### What this PR does
Add file maxSize validation support to MultiDropZone component.

1. Added support to maxSize prop and internal validation of size
2. Added formatBytes util to show more human readable information on file size

Solves:  
EMU-5671

Screenshot 2023-01-27 at 10 45 04

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/multiDropzone/index.test.tsx
+++ b/src/lib/components/multiDropzone/index.test.tsx
@@ -60,6 +60,21 @@ describe('MultiDropzone component', () => {
       expect(screen.getByText("Too many files.")).toBeVisible();
     });
 
+    it("should show max file size error message", async () => {
+      const screen = setup({maxSize: 10 });
+      const input = screen.getByTestId(inputTestId);
+      const bigFile = file;
+      Object.defineProperty(bigFile, 'size', { value: 1024 });
+
+      await act(async () => {
+        fireEvent.change(input, { target: { files: [bigFile] } });
+      });
+
+      expect(
+        screen.getByText("File is too large. It must be less than 10 Bytes.")
+      ).toBeInTheDocument();
+    });
+
     it("should show wrong filetype error message", async () => {
       const screen = setup({ accept: "document" });
       const input = screen.getByTestId(inputTestId);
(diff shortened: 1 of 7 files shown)


src/lib/util/formatBytes/index.ts
```diff
@@ -0,0 +1,13 @@
+const k = 1024;
+const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB'];
+
+export function formatBytes(bytes: number, decimals = 0): string {
```
diogomateus, line 4: Based on this stack overflow answer:
https://stackoverflow.com/questions/15900485/correct-way-to-convert-size-in-bytes-to-kb-mb-gb-in-javascript

Which is based on this script here, but was updated to be ES6:
https://web.archive.org/web/20120507054320/http://codeaid.net/javascript/convert-size-in-bytes-to-human-readable-format-(javascript)

src/lib/components/multiDropzone/index.test.tsx
```diff
@@ -60,6 +60,21 @@ describe('MultiDropzone component', () => {
...
     });
 
+    it("should show max file size error message", async () => {
+      const screen = setup({maxSize: 10 });
+      const input = screen.getByTestId(inputTestId);
+      const bigFile = file;
+      Object.defineProperty(bigFile, 'size', { value: 1024 });
+
+      await act(async () => {
+        fireEvent.change(input, { target: { files: [bigFile] } });
```
talo242, line 70: _This comment does not require any changes in this PR_
I think we'll have to update `testing-library` and add `user-event` pretty soon. As noted in the testing-library documentation, [using fireEvent is not recommended and instead we should use user-event](https://testing-library.com/docs/user-event/intro#differences-from-fireevent).
Can you create a follow up ticket for tackling this in the future?

talo242 (approved):
Left a comment regarding a follow up ticket, but from my point of view this PR is ready to be merged :shipit:

src/lib/components/multiDropzone/index.test.tsx
diogomateus: Noted and 100% agree!

PR #170, opened 2023-02-02
refactor(tests): [EMU-5691] Add user-event to tests configuration

diogomateus:
### What this PR does
Updates tests configuration and packages to have `react-hooks` and `user-event` packages available for testing.
This PR also updates the setup configuration to have userTest available, then making used of it and updating current tests to use user-event instead of fire-event.

### Why is this needed?
[Using fireEvent is not recommended and instead we should use user-event](https://testing-library.com/docs/user-event/intro#differences-from-fireevent).

Solves:  
EMU-5691

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge


kseha1080 (approved):

busypeoples (approved):

kimkwanka (approved):

src/lib/components/segmentedControl/index.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { fireEvent, render } from '@testing-library/react';
+import { render } from '../../util/testUtils';
```
talo242, line 1: This is a team style preference, we tried to use absolute paths whenever possible to keep things clear and nice:

```suggestion
import { render } from 'util/testUtils';
```

src/lib/components/multiDropzone/index.test.tsx
```diff
@@ -1,5 +1,4 @@
-import { fireEvent, render } from '@testing-library/react';
-import { act } from 'react-dom/test-utils';
+import { act, fireEvent, render } from '../../util/testUtils';
```
talo242, line 1: Same comment regarding absolute path.

src/lib/components/input/currency/index.test.tsx
```diff
@@ -1,4 +1,4 @@
-import { fireEvent, render } from '@testing-library/react';
+import { render } from '../../../util/testUtils';
```
talo242, line 1: And here

src/lib/components/autocompleteAddress/index.test.tsx
```diff
@@ -55,37 +55,41 @@ describe('AutocompleteAddress component', () => {
       }
     };
 
-    const screen = setup();
-    fireEvent.change(screen.getByTestId(inputTestId), {
+    const { getAllByTestId, getByTestId } = setup();
+
+    fireEvent.change(getByTestId(inputTestId), {
```
talo242, line 60: Is there a reason to leave `fireEvent` here instead of using `userEvent`?

src/lib/components/autocompleteAddress/index.test.tsx
```diff
@@ -55,37 +55,41 @@ describe('AutocompleteAddress component', () => {
...
       target: { value: 'Köpeniker' },
     });
 
-    const inputs = screen.getAllByTestId(inputTestId);
-    expect(inputs[0].getAttribute('value')).toBe('Köpeniker Strasse');
+    const inputs = getAllByTestId(inputTestId);
+
     expect(inputs.length).toEqual(5);
+    expect(inputs[0].getAttribute('value')).toBe("Köpeniker Strasse");
```
talo242, line 67: I would prefer to check if the text is visible on the screen. I.e. `screen.getByText('Köpeniker Strasse')`
What do you think?

talo242 (changes requested):

src/lib/components/autocompleteAddress/index.test.tsx
diogomateus: Makes sense, I'd suggest using the `getDisplayValue` instead:
`expect(getByDisplayValue("Köpeniker Strasse")).toBeVisible();`

src/lib/components/segmentedControl/index.test.tsx
diogomateus: Doesn't seem to work on dirty swan 😢 I need to check but maybe we're missing some alias configuration

src/lib/components/autocompleteAddress/index.test.tsx
busypeoples: How does `userEvent` differ from `fireEvent` technically?

busypeoples: We should keep in mind, that `autocomplete` in this case adds a listener to the input completely devoid of the React life-cycle.

diogomateus: As far as I know, userEvent is built on top of fireEvent but adds some "user like" goodies, like pointer up and down on click, checking the type of element being triggered and starting from that it has slightly different behaviour. 
You can check the implementation of here:
https://github.com/testing-library/user-event/blob/5feaa942f46bb37d96c2f2fbeb4b33e8beff75ad/src/click.js#L87-L103

Regarding the autocomplete, I agree it's something we should be aware of and, IMO, even take a second look to see if it's possible to update it to be part of the react cycle as you mentioned

src/lib/components/segmentedControl/index.test.tsx
talo242: Ohhh yeah, we probably need to update the TS config file. Ok all good on this PR then.

src/lib/components/autocompleteAddress/index.test.tsx
talo242: > We should keep in mind, that autocomplete in this case adds a listener to the input completely devoid of the React life-cycle.

True, I forgot about this custom implementation. Ok, let's stick with `userEvent`.
Adding up to what Diogo said regarding the differences, `userEvent` abstracts a little bit more the event functionalities and doesn't allow you to mess with internals. That gives you an environment more similar to what a user would experience.

talo242 (approved):

PR #172, opened 2023-02-02
refactor(multidropzone): [EMU-5692] Add file type and mime type definitions

diogomateus:
### What this PR does
Add file type and mime type definitions to a shared types folder.
This PR also updates MultiDropzone component to take advantage of the file type/mime-type enums based on a mozilla doc configuration instead of manually setting the mime-type.

Solves:  
EMU-5692

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge


src/lib/types/file-type.ts
```diff
@@ -0,0 +1,83 @@
+// Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+export enum FileMimeTypes {
```
diogomateus, line 2: This is being added here but is probably worth thinking if it should be either smaller and for the specificity of the MultiDropzone component or even created and imported in another repo like https://github.com/getPopsure/public-models

talo242: I'd argue that we could start with a smaller list and add more types as we need them.
Let's see what the other teammates have to say about this.

kseha1080: I think we can limit it to the specific file types that we / ops team want then just add the ones we want later. Since the defined type is specific to this upload component, I think it makes that it lives here and then we can just export it. wdyt?

diogomateus: That makes sense to me. I just updated the code to use a smaller file that defined inside of MultiDropzone types.

Some notes on this update:
- I made `type` on UploadedFile not required as it is not being used anymore but I don't want to introduce any breaking changes,
- The only case where we might need additional FileTypes is if we add a custom `accept` as the component accepts any filetypes by default.

src/lib/components/multiDropzone/types.ts
```diff
@@ -1,17 +1,30 @@
...
+  pdf = "application/pdf",
+  png = "image/png",
+  tif = "image/tiff",
+  tiff = "image/tiff",
+  webp = "image/webp",
+}
 
-export const DOCUMENT_FILES = ['doc', 'docx', 'pdf'];
-export const IMAGE_FILES = ['heic', 'bmp', 'jpeg', 'jpg', 'png'];
+export type FileType = keyof typeof FileMimeTypes;
```
kseha1080, line 17: It'd be useful to export this from the lib, wdyt?

diogomateus: Agree, I think we should start exporting more type definitions from the DS as they can be handy on implementations

diogomateus: Updated here: 
https://github.com/getPopsure/dirty-swan/pull/172/commits/39f7f4a96c14b342205173d6335a3431aafe49c5

src/lib/index.tsx
```diff
@@ -33,6 +38,13 @@ import {
 import SegmentedControl from './components/segmentedControl';
 import Markdown from './components/markdown';
 
+export type {
+  FileType,
+  MultiDropzoneProps,
+  UploadedFile,
+  UploadStatus
+};
```
kseha1080, line 46: ✌️

kseha1080 (approved):

talo242 (approved):
Looking good ✨

PR #174, opened 2023-02-03
refactor(tsconfig): [EMU-5705] Add baseUrl to tsconfig allowing relative paths

diogomateus:
## What this PR does
Adds baseUrl to tsconfig allowing relative paths. This allows to use
```
import foo from "lib/utils/foo";
```
instead of

```
import bar from "../../../utils/bar";
```

Note: Not possible to add just `from "utils/foo"` as this is not supported by Create React App.

Solves:  
EMU-5705

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/autocompleteAddress/index.test.tsx
+++ b/src/lib/components/autocompleteAddress/index.test.tsx
@@ -1,5 +1,5 @@
 import { Address } from '@popsure/public-models';
-import { fireEvent, render } from '../../util/testUtils';
+import { fireEvent, render } from 'lib/util/testUtils';
 
 import AutoCompleteAddress from '.';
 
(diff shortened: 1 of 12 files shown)


kimkwanka:
Have you tested a local build to see if the package is still usable once it has been compiled to JS? I'm not sure if the relative paths will break the build in this case, but in my experience, it is definitely something to look out for. [This notion page](https://www.notion.so/featherinsurance/Working-with-NPM-packages-25cde1ab5d4c43d19c456ac9cd36f971#e56c16d21c4e4825a4aaf5262171945a) shows you how you can use your local build as a package in another project.

PR #178, opened 2023-02-07
refactor(comparison table): [EMU-5713] Add TableButton to TableRowHeader component

diogomateus:
### What this PR does

1. Adds TableButton component and exports it
2. Hides icon span when icon is not passed to TableRowHeader
3. Updates TableRowHeader to use TableButton instead of TableInfoButton
4. Adds tests to TableButton and TableRowHeader component
5. Adds an example of TableRowHeader and TableButton components to ComparisonTable story, as shown on the screenshot below

Screenshot 2023-02-07 at 13 45 15

### Why is this needed?

_Please include additional context of why this PR is necessary._

Solves:  
[EMU-5713](https://linear.app/feather-insurance/issue/EMU-5713/update-comparisontable-component-styles)

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/comparisonTable/components/TableButton/index.tsx
+++ b/src/lib/components/comparisonTable/components/TableButton/index.tsx
@@ -10,15 +10,13 @@ const TableButton: React.FC<Props> = ({
   onClick,
   className = '',
 }) => (
-  <span
+  <button
     className={`${styles.button} ${className}`}
     data-testid="ds-table-button"
     onClick={onClick}
-    role="button"
-    tabIndex={0}
   >
     {children}
-  </span>
+  </button>
 );
 
 export default TableButton;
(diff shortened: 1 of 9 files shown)


src/lib/components/comparisonTable/components/TableButton/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+  onClick,
+  className = '',
+}) => (
+  <span
+    className={`${styles.button} ${className}`}
+    data-testid="ds-table-button"
+    onClick={onClick}
+    role="button"
+    tabIndex={0}
+  >
```
talo242, line 19: Any reason to use `span` instead of a `button` tag?
I'm not a big fan of adding button capabilities to non interactive elements, if possible I'd suggest switching this to a regular `button` and fix styles if necessary.
What do you think?

diogomateus: Not really, I was following the other table button example but if thinking critically about it there's no reason for this to be this way. Just updated, also added an outline color based on the button outline for accessibility purposes.

talo242 (approved):
🙌🏼

PR #181, opened 2023-02-08
fix(multidropzone): [noticket] Remove unused uploaded filetype and export types

diogomateus:
### What this PR does
Removed uploaded as this value is not used anymore.
This also adds MultiDropzone types to be exported from lib index file.

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/multiDropzone/types.ts
+++ b/src/lib/components/multiDropzone/types.ts
@@ -24,7 +24,6 @@ export const IMAGE_FILES: FileType[] = ['heic', 'bmp', 'jpeg', 'jpg', 'png'];
 export interface UploadedFile {
   id: string;
   name: string;
-  type?: FileType | string;
   previewUrl?: string;
   progress: number;
   error?: string;
--- a/src/lib/components/multiDropzone/types.ts
+++ b/src/lib/components/multiDropzone/types.ts
@@ -24,7 +24,7 @@ export const IMAGE_FILES: FileType[] = ['heic', 'bmp', 'jpeg', 'jpg', 'png'];
 export interface UploadedFile {
   id: string;
   name: string;
-  type?: FileType;
+  type?: FileType | string;
   previewUrl?: string;
   progress: number;
   error?: string;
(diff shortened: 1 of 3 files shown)


src/lib/components/multiDropzone/types.ts
```diff
@@ -24,7 +24,7 @@ export const IMAGE_FILES: FileType[] = ['heic', 'bmp', 'jpeg', 'jpg', 'png'];
 export interface UploadedFile {
   id: string;
   name: string;
-  type?: FileType;
+  type?: FileType | string;
```
busypeoples, line 27: I think you need to fix this issue on the consumer side otherwise you will lose the benefits of defining `FileType`.
This means all applications using the `MultDropZone` need to adapt their types.

busypeoples (changes requested):

src/lib/components/multiDropzone/types.ts
busypeoples: for example in `app` you would need to change the `type` from string to `FileType` etc.
that way you can ensure that the defined upload file type confirms to the `FileType` in ds.

```
export interface UploadedFile {
  id: string;
  name: string;
  type: string;
  progress: number;
  error?: string;
  token?: string;
  previewUrl?: string;
}
```

talo242: I agree with @busypeoples, we have to make types stricter on the consumer projects. I think we are already exporting `FileType` from DS, so we can reuse that type on the consumer side.

diogomateus: I was trying not to introduce any breaking changes. If we should update on the consumer side, I recommend even removing this type as we are not using it anymore (it was used before to show an icon according to the type of file).

I can update this PR to remove it and fix it on app side. As it's already optional, it is an improvement but we don't need to wait to have this released to fix it.

talo242 (approved):
@diogomateus if the type was being used only for the icon then it makes sense to remove the prop.
Thanks for checking into this!

PR #182, opened 2023-02-13
fix(table button): [noticket] Fix types and underlined styles

diogomateus:
### What this PR does
Adds children type to TableButton component explicitely as in some versions of react, children doesn't exist on React.FC type causing a typescript error.
Fixes underlined styles for TableButtonHeader component when multiline. Still needs to be kept as a border because on non text underline behaviour.

**Before**
Screenshot 2023-02-13 at 09 04 03

**After**
Screenshot 2023-02-13 at 09 03 49

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/comparisonTable/components/TableButton/style.module.scss
+++ b/src/lib/components/comparisonTable/components/TableButton/style.module.scss
@@ -2,17 +2,26 @@
 
 .button {
   background-color: transparent;
-  border-bottom: 2px dashed $ds-grey-600;
   color: $ds-grey-700;
   cursor: pointer;
+  margin: 2px 0;
   padding: 2px;
-  transition: 0.3s ease;
-  transition-property: color, border-color;
+  text-align: left;
+  transition: color 0.3s ease;
+
+  span {
+    border-bottom: 2px dashed $ds-grey-600;
+    display: inline;
+    transition: border-color 0.3s ease;
+  }
 
   &:hover,
   &:focus {
-    border-color: $ds-primary-500;
     color: $ds-primary-500;
     outline-color: $ds-primary-500;
+    
+    span {
+      border-color: $ds-primary-500;
+    }
   }
 }
--- a/src/lib/components/comparisonTable/components/TableButton/style.module.scss
+++ b/src/lib/components/comparisonTable/components/TableButton/style.module.scss
@@ -6,7 +6,6 @@
   cursor: pointer;
   margin: 2px 0;
   padding: 2px;
-  text-align: left;
   transition: color 0.3s ease;
 
   span {
--- a/src/lib/components/comparisonTable/index.stories.mdx
+++ b/src/lib/components/comparisonTable/index.stories.mdx
@@ -85,21 +85,33 @@ export const ComparisonTableWithData = () => {
           id: 4,
           key: 'imdb',
           label: 'IMDB rating',
+          render: (value) => (
+            <TableButton onClick={() => {}}>
+              {value}
+            </TableButton>
+          ),
         },
         {
           id: 3,
           key: 'rating',
-          label: <TableRowHeader label="Our Rating" onClickInfo={() => {}} />,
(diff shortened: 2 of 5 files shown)


src/lib/components/comparisonTable/components/TableButton/style.module.scss
```diff
@@ -2,17 +2,26 @@
...
 .button {
   background-color: transparent;
-  border-bottom: 2px dashed $ds-grey-600;
   color: $ds-grey-700;
   cursor: pointer;
+  margin: 2px 0;
   padding: 2px;
-  transition: 0.3s ease;
-  transition-property: color, border-color;
+  text-align: left;
```
kimkwanka, line 9: Hardcoding a left alignment for text is not a good idea as it only really works if the button is used as the table row header. If it's used to turn some other text into a clickable unit, it will lead to inconsistencies as the default is centered text:
image

src/lib/components/comparisonTable/index.stories.mdx
```diff
@@ -77,21 +77,34 @@ export const ComparisonTableWithData = () => {
...
+            <TableButton onClick={() => {}}>
+              {value}
+            </TableButton>
+          ),
         },
         {
           id: 3,
           key: 'rating',
-          label: <TableRowHeader label="Our Rating" onClickInfo={console.log} />,
+          label: <TableRowHeader label="Our rating of movies found online on IMDB lists" onClickInfo={() => {}} />,
```
kimkwanka, line 89: If you change the label here, please also update it in the "pseudo code" that's shown for consistency.

src/lib/components/comparisonTable/index.stories.mdx
```diff
@@ -77,21 +77,34 @@ export const ComparisonTableWithData = () => {
...
           render: (value) => (
-            <TableButton className="jc-center" onClick={console.log}>
-              <TableRating type="star" rating={value} />
+            <span>
+            <TableButton onClick={() => {}}>
+              <TableRating type="zap" rating={value} />
```
kimkwanka, line 93: Not sure why the stars got turned into zaps here.

kimkwanka (changes requested):
Changes requested for the comments.

src/lib/components/comparisonTable/index.stories.mdx
kimkwanka: image

src/lib/components/comparisonTable/components/TableButton/style.module.scss
diogomateus: Thanks for this catch! Completely missed the data table/content with multi line. Kept center as default and aligning left on table row headers only, as it should.

src/lib/components/comparisonTable/index.stories.mdx
```diff
@@ -77,21 +77,32 @@ export const ComparisonTableWithData = () => {
...
         {
           id: 3,
           key: 'rating',
-          label: <TableRowHeader label="Our Rating" onClickInfo={console.log} />,
+          label: <TableRowHeader label="Our rating of movies found online on IMDB lists" onClickInfo={() => {}} />,
           render: (value) => (
-            <TableButton className="jc-center" onClick={console.log}>
-              <TableRating type="star" rating={value} />
+            <TableButton onClick={() => {}}>
+              <TableRating type="stars" rating={value} />
```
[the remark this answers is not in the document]
kimkwanka: Apart from that everything looks good 👍 ✨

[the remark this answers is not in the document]
diogomateus: Ah! Mdx + typescript isn't showing this and since it has the default value it all looked good from a testing perspective, just updated and will look into a way to make sure typescript, eslint and mdx all behave properly 🙇

kimkwanka (approved):

PR #184, opened 2023-02-15
refactor(uuid): [noticket]: Replace uuid package with generateId util

diogomateus:
### What this PR does
Replaces uuid package with a custom and simpler generateId util.

### Why is this needed?
uuid package can't be tree-shaked and was adding a giant additional size to repos where it was being imported.
Usually, using random for random ids is not the most efficient way but since this ids are not persistent and used for runtime only (generate label/input ids, temporary ids for error entries) this downgrade from uuid might make sense.

Named the util generateId so it doesn't suggest it is a fully and standardised uuid and replaces the function.
The id generated is based on a random number that is then parsed into an hex string for each of the characters.

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/util/generateId/index.test.ts
+++ b/src/lib/util/generateId/index.test.ts
@@ -1,0 +1,10 @@
+import generateId from '.';
+
+describe('generateId', () => {
+  it('should return two different ids', () => {
+    const firstId = generateId();
+    const secondId = generateId();
+
+    expect(firstId).not.toEqual(secondId);
+  });
+});
--- a/src/lib/util/generateId/index.ts
+++ b/src/lib/util/generateId/index.ts
@@ -1,0 +1,7 @@
+function generateId(): string {  
+  return 'xxxx-xxxx-xxx-xxxx'.replace(/[x]/g, () => (
+    Math.floor(Math.random() * 16).toString(16)
+  ));  
+}
+
+export default generateId;
--- a/src/lib/util/generateId/index.ts
+++ b/src/lib/util/generateId/index.ts
@@ -1,7 +1,11 @@
 function generateId(): string {  
-  return 'xxxx-xxxx-xxx-xxxx'.replace(/[x]/g, () => (
-    Math.floor(Math.random() * 16).toString(16)
-  ));  
+  if (!window?.crypto?.randomUUID) {
+    return 'xxxx-xxxx-xxx-xxxx'.replace(/[x]/g, () => (
+      Math.floor(Math.random() * 16).toString(16)
+    ));
+  }
+
+  return crypto.randomUUID();
 }
 
 export default generateId;
--- a/src/lib/util/generateId/index.ts
+++ b/src/lib/util/generateId/index.ts
@@ -1,11 +1,15 @@
-function generateId(): string {  
-  if (!window?.crypto?.randomUUID) {
-    return 'xxxx-xxxx-xxx-xxxx'.replace(/[x]/g, () => (
-      Math.floor(Math.random() * 16).toString(16)
-    ));
+function generateId(): string {
+  if (window?.crypto?.getRandomValues) {
+    return (
+      `${[1e7]}-${1e3}-${4e3}-${8e3}-${1e11}`).replace(/[018]/g,
+      (c: any) =>
+        (((c ^ window.crypto.getRandomValues(new Uint8Array(1))[0]) & 15) >> c / 4).toString(16)
+    );
   }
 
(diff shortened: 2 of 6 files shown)


diogomateus:
@everyone I'm very open to other proposals regarding this refactoring.

src/lib/util/generateId/index.test.ts
```diff
@@ -0,0 +1,10 @@
+import generateId from '.';
+
+describe('generateId', () => {
+  it('should return two different ids', () => {
+    const firstId = generateId();
+    const secondId = generateId();
+
+    expect(firstId).not.toEqual(secondId);
+  });
```
talo242, line 9: Not sure if it'd be worth it to write regression tests for the format `xxxx-xxxx-xxx-xxxx` and/or length of the string? (only if it's relevant)

diogomateus: You regression as a way to make sure it's compatible with previous usages?

kimkwanka:
Instead of cooking up our own solution here which, in theory, could still produce identical ids, would maybe using https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID be an alternative here? It looks like browser support is solid

src/lib/util/generateId/index.ts
```diff
@@ -0,0 +1,11 @@
+function generateId(): string {  
+  if (!window?.crypto?.randomUUID) {
```
diogomateus, line 2: Thanks for that suggestion, @kimkwanka!

Added a fallback for the edge cases although support is over 90% of browsers, wdyt?

kimkwanka: Sounds great 👍

kimkwanka (approved):

diogomateus:
Seems like `randomUUID` runs into a lot of issues and is only available in [secure contexts](https://github.com/WICG/uuid/issues/23) meaning we might run into some issues using it. On top of that it's only available on node v19 and we're running on v16 which means it triggers errors when building storybook.

Updated it to use this function by broofa that is one of the contributors to the uuid library and, I believe, part of the team that implemented randomUUID.

Still defaulting to Math random for places where we might run into an issue. Testing should be one those cases.

kimkwanka:
> Seems like `randomUUID` runs into a lot of issues and is only available in [secure contexts](https://github.com/WICG/uuid/issues/23) meaning we might run into some issues using it. On top of that it's only available on node v19 and we're running on v16 which means it triggers errors when building storybook.
> 
> Updated it to [use this function by broofa ](https://stackoverflow.com/a/2117523)that is one of the contributors to the uuid library and, I believe, part of the team that implemented randomUUID.
> 
> Still defaulting to Math random for places where we might run into an issue. Testing should be one those cases.

That's a shame, but I guess that's all that can be done at this point. Thanks 👍

kimkwanka (approved):

diogomateus:
> > Seems like `randomUUID` runs into a lot of issues and is only available in [secure contexts](https://github.com/WICG/uuid/issues/23) meaning we might run into some issues using it. On top of that it's only available on node v19 and we're running on v16 which means it triggers errors when building storybook.
> > Updated it to [use this function by broofa ](https://stackoverflow.com/a/2117523)that is one of the contributors to the uuid library and, I believe, part of the team that implemented randomUUID.
> > Still defaulting to Math random for places where we might run into an issue. Testing should be one those cases.
> 
> That's a shame, but I guess that's all that can be done at this point. Thanks 👍

100%, pretty annoying that we have to go through such hassle just to create an ID when there's even a function already for it 😅

broofa:
> uuid package can't be tree-shaked

Can you elaborate on this?  The `uuid` module was refactored a while back precisely to facillitate tree-shaking, so this comment is  a little surprising.

kimkwanka:
> > uuid package can't be tree-shaked
> 
> Can you elaborate on this? The `uuid` module was refactored a while back precisely to facillitate tree-shaking, so this comment is a little surprising.

@diogomateus

diogomateus:
> > uuid package can't be tree-shaked
> 
> Can you elaborate on this? The `uuid` module was refactored a while back precisely to facillitate tree-shaking, so this comment is a little surprising.

After digging up a bit, I think this might be related to the fact that we were using the v0.8.3 instead of the most recent one where the minified UMD build was removed so there shouldn't be an issue with it. I'll take a look at replicating it again later.

PR #185, opened 2023-02-17
fix(generateId): Add additional check for undefined window

diogomateus:
### What this PR does
Adds an additional check for undefined window that was left out of https://github.com/getPopsure/dirty-swan/pull/184 by mistake. This doesn't run into any issues on dirty-swan checks but will run into server side run projects like the ones using nextjs.

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge


diogomateus:
We could look into adding a way of testing emulation of SSR so this is caught on dirty-swan side.

kimkwanka (approved):

PR #186, opened 2023-02-20
fix(comparison table): Update default header width size

diogomateus:
### What this PR does
This PR updates the default header width size from 0 to 1400.
This doesn't break any layout as there are resize observers updating to the final size. However, this improves the initial rendering making the content not glitch/jump since 1400 is closer to 100% than 0 and content shouldn't grow over 100% of viewport anyway. The purpose of this is to fix flaky tests. Not using window width since it breaks on SSR.

### Checklist:

- [X] I reviewed my own code
- [X] The changes align with the designs I received  
- [X] I have attached screenshot(s), video(s) or gif(s) showing that the solution is working as expected  
- [X] I have updated the task(s) status on Linear
- [X] All new media is optimized (images, gifs, videos)

### Browser support

- [X] Firefox
- [X] Chrome
- [X] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/comparisonTable/components/Row/style.module.scss
+++ b/src/lib/components/comparisonTable/components/Row/style.module.scss
@@ -8,7 +8,7 @@ $cell-min-width: var(--cellWidth, 256px);
 
   & > div {
     width: var(--tableWidth);
-    max-width: var(--tableWidth);
+    max-width: calc(100vw - 32px);
 
     &:nth-child(n + 3) {
       margin: 0;
--- a/src/lib/components/comparisonTable/components/Row/style.module.scss
+++ b/src/lib/components/comparisonTable/components/Row/style.module.scss
@@ -8,6 +8,7 @@ $cell-min-width: var(--cellWidth, 256px);
 
   & > div {
     width: var(--tableWidth);
+    // 100% of viewport width minus 16px horizontal padding
     max-width: calc(100vw - 32px);
 
     &:nth-child(n + 3) {
--- a/src/lib/components/comparisonTable/hooks/useComparisonTable.ts
+++ b/src/lib/components/comparisonTable/hooks/useComparisonTable.ts
@@ -5,7 +5,7 @@ import { ArrowValues } from '../components/TableArrows';
 
 export const useComparisonTable = () => {
   const [showMore, setShowMore] = useState<boolean>(false);
-  const [headerWidth, setHeaderWidth] = useState(0);
+  const [headerWidth, setHeaderWidth] = useState(1400);
 
   const [selectedTabIndex, setSelectedTabIndex] = useState(0);
 


src/lib/components/comparisonTable/hooks/useComparisonTable.ts
```diff
@@ -5,7 +5,7 @@ import { ArrowValues } from '../components/TableArrows';
 
 export const useComparisonTable = () => {
   const [showMore, setShowMore] = useState<boolean>(false);
-  const [headerWidth, setHeaderWidth] = useState(0);
+  const [headerWidth, setHeaderWidth] = useState(1400);
```
kimkwanka, line 8: Just to get some context here, how did you get to the value `1400`?

diogomateus: 1400 seemed like a good compromise value as to what a container can be, although it is a pretty artificial and random number.

In a way, it doesn't really matter, as long as it isn't `0` or a really small value since the width of cells will always show a minimum value (tablewidth is the headerWidth value here):
`max-width: min(var(--firstColumnWidth, 288px), calc(var(--tableWidth) / 2));`

However, the goal here is moving away from zero to avoid the width changes jump that are visible. It could even be 2000 or bigger.

talo242: If I understand correctly, the flakiness comes from `headerWidth` starting with `0` and then getting resized inside the `useEffect` on line #26?

src/lib/components/comparisonTable/components/Row/style.module.scss
```diff
@@ -8,7 +8,7 @@ $cell-min-width: var(--cellWidth, 256px);
 
   & > div {
     width: var(--tableWidth);
-    max-width: var(--tableWidth);
+    max-width: calc(100vw - 32px);
```
talo242, line 11: One question, what is the 32px here? is it some padding?
It may be good to add a comment explaining.

src/lib/components/comparisonTable/hooks/useComparisonTable.ts
diogomateus: Exactly, that is what It seems to be causing the flakiness, resizeObserver and useEffect take some time to apply the final width size.

talo242: It seems like a good enough solution for me. The `headerWidth` will be set to the correct value once the component finishes mounting.
I'm just wondering if this will indeed prevent the flakiness. Is there a case when the `headerWidth` starts bigger than it should and then the `useEffect` squeeze it back to the correct size?

diogomateus: From what I was testing it seems like that never happens because of this line setting the max-width of cells:
`max-width: min(var(--firstColumnWidth, 288px), calc(var(--tableWidth) / 2));`

This makes the max width be the minimum value of either `288px` or `tablewidth / 2` which only fires if table width is a really low value like zero.

The only issue is the scrolling header on mobile that I'm setting the max-width to be `calc(100vw - 32px)`.

kimkwanka: Just gave your fix a spin and it seems to not break anything that I was able to test, fingers crossed that this will indeed fix the flakiness 🤞 😄

diogomateus: Same here, really hope this is the cause! 🤞🤞

talo242 (approved):
🙌🏼

PR #187, opened 2023-02-21
fix(test-build-storybook): Set fixed version of node

diogomateus:
### What this PR does
Set fixed `v16.10.0` version of node when running `test-build-storybook`.

### Why is this needed?
Current PRs were[ failing when building storybook](https://github.com/getPopsure/dirty-swan/actions/runs/4231087840/jobs/7349172627) since github actions were somehow using node v17 instead of the ones used previously. No changes were made before to trigger this error.

### Checklist:

- [X] I reviewed my own code
- [X] The changes align with the designs I received  

### Browser support

- [X] Firefox
- [X] Chrome
- [X] Safari
- [X] Edge

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -34,6 +34,12 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
+      - name: Install NodeJS
+        uses: actions/setup-node@v3
+        with:
+          node-version: '16.10.0'
+          cache: 'yarn'
+
       - name: yarn install & build-storybook
         run: |
           yarn install


vincentaudoire (approved):

diogomateus:
I wasn't able to find the root cause for this. Should we add this to other scenarios as well where we want the node version to be set instead of relying on GH Actions?

talo242:
> I wasn't able to find the root cause for this. Should we add this to other scenarios as well where we want the node version to be set instead of relying on GH Actions?

Do you mean on the other GH Actions in this project? I don't see a problem with that, as long as wee keep updating Node every once in a while.

diogomateus:
> > I wasn't able to find the root cause for this. Should we add this to other scenarios as well where we want the node version to be set instead of relying on GH Actions?
> 
> Do you mean on the other GH Actions in this project? I don't see a problem with that, as long as wee keep updating Node every once in a while.

Yes, that's what I meant. I just created another PR to update [it here](https://github.com/getPopsure/dirty-swan/pull/188) as it is also breaking deploy. 
Release already defines a version.

PR #189, opened 2023-02-24
refactor(fonts): switch from import to font-face

busypeoples:
### What this PR does

Switch from @import to @font-face when loading fonts to prevent blocking

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/scss/private/base/_typography.scss
+++ b/src/lib/scss/private/base/_typography.scss
@@ -5,8 +5,25 @@
 
 @font-face {
   font-family: 'Merriweather';
+  font-style: normal;
+  font-weight: 700;
   font-display: swap;
-  src: url('https://fonts.googleapis.com/css?family=Merriweather:700&display=swap');
+  src: url(https://fonts.gstatic.com/s/merriweather/v30/u-4n0qyriQwlOrhSvowK_l52xwNZVsf6hPvhPUWH.woff2)
+    format('woff2');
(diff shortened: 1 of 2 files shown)


src/lib/scss/private/base/_typography.scss
```diff
@@ -3,7 +3,11 @@
 @use '../../public/grid' as *;
 @use '../../public/font' as *;
 
-@import url('https://fonts.googleapis.com/css?family=Merriweather:700&display=swap');
+@font-face {
+  font-family: 'Merriweather';
+  font-display: swap;
+  src: url('https://fonts.googleapis.com/css?family=Merriweather:700&display=swap');
```
vincentaudoire, line 9: You're still loading the css file - don't you need to load the actual font file here? 
`  src: url(https://fonts.gstatic.com/s/merriweather/v30/u-4n0qyriQwlOrhSvowK_l52xwNZWMf6hPvhPQ.woff2) format('woff2');` (as taken from https://fonts.googleapis.com/css2?family=Merriweather:wght@700&display=swap).

I would basically copy over the two value `latin-ext` and `latin` over (or maybe we don't even need `latin-ext`

vincentaudoire (changes requested):

vincentaudoire (approved):

PR #192, opened 2023-02-24
fix(modal): Fix margin and overlay area

diogomateus:
### What this PR does
Fixes overlay area so that all clicks outside body close the modal (including margin). Also fixes if dismissable, making it not closable on overlay click.
Adds a bigger margin for bigger modals, so that they don't show over navbar. This only impacts modals with height bigger than window height.

**Note:** While this should be a zIndex fix on navbar side, it's an easy way to fix without impacting other content and adding a nice margin to content. Should still be checked on the other side of the application.

**Before:**
Screenshot 2023-02-24 at 11 55 54

**After:**
Screenshot 2023-02-24 at 15 25 07

### Why is this needed?

_Please include additional context of why this PR is necessary._

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/modal/bottomModal/index.tsx
+++ b/src/lib/components/modal/bottomModal/index.tsx
@@ -15,7 +15,12 @@ export default ({
   dismissible = true,
 }: Props) => {
   const [containerXOffset, setContainerXOffset] = useState(0);
-  const { isClosing, handleContainerClick, handleOnClose } = useOnClose(
+  const {
+    isClosing,
+    handleContainerClick,
+    handleOnClose,
+    handleOnOverlayClick
+   } = useOnClose(
     onClose,
     isOpen,
     dismissible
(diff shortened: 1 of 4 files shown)


kimkwanka:
It seems like this breaks the "non-dismissible" variant as it now closes the modal when clicking the container margin. (Although that variant had issues to begin with, closing when you click the area left or right of the modal.)

diogomateus:
> It seems like this breaks the "non-dismissible" variant as it now closes the modal when clicking the container margin. (Although that variant had issues to begin with, closing when you click the area left or right of the modal.)

Just fixed it, seems like it wasn't doing anything for dismissable on overlay click. I'll try to get in some testing in here as well :)

kimkwanka (approved):

PR #196, opened 2023-03-27
fix(currency-input): [EMU-5966] Fix currencyInput selectionStart after value format

diogomateus:
### What this PR does
Fix currencyInput selectionStart after value format, making the cursor stay in the expected position instead of going to the end of the input.

Solves:  
EMU-5966

https://user-images.githubusercontent.com/4015038/227987170-270f8522-dadb-4956-ab79-a21393561529.mov

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [ ] Edge


ruizhen88 (approved):
Great test! 🧪

PR #206, opened 2023-03-30
Update calendar to close when clicking outside

busypeoples:
### What this PR does

Ensure that the calendar closes when clicking outside

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/dateSelector/index.tsx
+++ b/src/lib/components/dateSelector/index.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState, useEffect, useRef } from 'react';
 import dayjs from 'dayjs';
 
 import localeData from 'dayjs/plugin/localeData';
@@ -97,6 +97,8 @@ const DateSelector = ({
   );
   const [openCalendar, setOpenCalendar] = useState(false);
 
+  const calendarContainerRef = useRef<HTMLDivElement | null>(null);
+
   const calendarDefaultDate =
     dayjs().year() >= yearBoundaries.min && dayjs().year() <= yearBoundaries.max
       ? dayjs().toDate()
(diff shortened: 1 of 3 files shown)


src/lib/components/dateSelector/index.tsx
```diff
@@ -111,6 +113,22 @@ const DateSelector = ({
...
+        setOpenCalendar(false);
+      }
+    };
+    document.addEventListener('mousedown', handleOutsideClick);
+
+    return () => {
+      document.removeEventListener('mousedown', handleOutsideClick);
+    };
+  }, [calendarContainerRef, setOpenCalendar]);
+
```
diogomateus, line 131: WDYT about moving this to a the utils or hooks folder? Seems like something that might be useful for other cases. We could even export it and take advantage of that in app/website/others repos

ruizhen88: Was going to make the same comment!

busypeoples: Makes sense.

ruizhen88 (approved):

src/lib/components/dateSelector/index.tsx
diogomateus: We could even move that hook to dirty-swan and import it on website, less duplicated code 🧹 🧹

busypeoples: Thanks @ruizhen88 @diogomateus for the feedback. Will check and then update.

talo242 (approved):
Such a nice contribution 🙌🏼

kimkwanka (approved):

src/lib/components/dateSelector/index.tsx
busypeoples: Ok converted the onClickOutside into a hook @diogomateus @ruizhen88

diogomateus: Looks good! Should we add some tests as well?

busypeoples: Ok, added some test @diogomateus

PR #208, opened 2023-03-31
feat(button): [RVN-212] Add spacing before outline in focus state

mbaguiar:
### What this PR does

* Adds space before outline in focus state of button

#### After
Screenshot 2023-03-31 at 10 31 28

#### Before
Screenshot 2023-03-31 at 10 54 40

Solves:
RVN-212

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Edge

### DIFF
--- a/src/lib/scss/private/components/_buttons.scss
+++ b/src/lib/scss/private/components/_buttons.scss
@@ -59,7 +59,8 @@
 
     &:focus {
       outline: none;
-      box-shadow: 0 0 0 2px rgba($color: $ds-primary-500, $alpha: 0.5);
+      box-shadow: 0 0 0 1px white,
+        0 0 0 3px rgba($color: $ds-primary-500, $alpha: 0.5);
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/lib/scss/private/components/_buttons.scss
```diff
@@ -59,7 +59,8 @@
 
     &:focus {
       outline: none;
-      box-shadow: 0 0 0 2px rgba($color: $ds-primary-500, $alpha: 0.5);
+      box-shadow: 0 0 0 1px white,
```
mbaguiar, line 62: I couldn't make the outline offset color to inherit the background color. I also checked and this is the default for the tailwindcss classes that do this. If anyone knows another way, let me know

mbaguiar: Transparent doesn't work, because it shows the color of the outline. Basically what happens here is we have one outline on top of the other to mimic an offset

diogomateus (approved):

busypeoples (approved):

PR #209, opened 2023-03-31
fix (table-info-button): [EMU-6047] Make TableInfoButton focusable

diogomateus:
### What this PR does

Make TableInfoButton focusable.

Solves:
EMU-6047

Non focused state:
Screenshot 2023-03-31 at 10 33 44

Focused state:
Screenshot 2023-03-31 at 10 33 42

### Checklist:

- [x] I reviewed my own code
- [x] The changes align with the designs I received  
  Or give a reason why this does not apply:
- [x] I have attached screenshot(s), video(s) or gif(s) showing that the solution is working as expected  
  Or give a reason why this does not apply:
- [x] I have updated the task(s) status on Linear
- [x] All new media is optimized (images, gifs, videos)

### Browser support

My code works in the following browsers:

- [x] Firefox
- [x] Chrome
- [x] Safari
- [ ] Edge

PR #210, opened 2023-03-31
fix(table-info-button): [EMU-6047] Make TableInfoButton click happen on enter

diogomateus:
### What this PR does

Related to https://github.com/getPopsure/dirty-swan/pull/209. This made the component focusable but since it's not a default browser button, it doesn't allow to trigger using keyboarding only which is a far from optimal user experience.

Solves:
EMU-6047

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/comparisonTable/components/TableInfoButton/index.tsx
+++ b/src/lib/components/comparisonTable/components/TableInfoButton/index.tsx
@@ -12,7 +12,9 @@ const TableInfoButton = ({
     className={`p-btn--secondary ${styles.button} ${className}`}
     onClick={onClick}
     onKeyDown={(e) => {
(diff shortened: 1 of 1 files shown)


kseha1080 (approved):

kimkwanka (approved):
I wonder if we should also support SPACE here, even if not strictly required by WCAG 2.0. Also maybe a `stopPropagation()` call should be added, just to be safe. https://www.w3.org/WAI/GL/wiki/Making_actions_keyboard_accessible_by_using_keyboard_event_handlers_with_WAI-ARIA_controls

busypeoples (approved):

PR #213, opened 2023-04-12
fix(comparison-table): Prevent scroll to update tab index on above mobile breakpoints

diogomateus:
### What this PR does

This prevents scrolling on table to update tab index on above mobile breakpoints. This means that on resize, we always have a same consistent result. This is specially important for visual regression testing as resizing the page would show the current tab randomly.

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/comparisonTable/hooks/useComparisonTable.ts
+++ b/src/lib/components/comparisonTable/hooks/useComparisonTable.ts
@@ -56,12 +56,15 @@ export const useComparisonTable = () => {
       return;
     }
 
+    const headerWidth = headerRef.current.getBoundingClientRect().width;
+
     const currentTabIndex = Math.round(
-      headerRef.current.scrollLeft /
-        headerRef.current.getBoundingClientRect().width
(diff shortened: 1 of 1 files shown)


src/lib/components/comparisonTable/hooks/useComparisonTable.ts
```diff
@@ -56,12 +56,15 @@ export const useComparisonTable = () => {
...
+    const headerWidth = headerRef.current.getBoundingClientRect().width;
+
     const currentTabIndex = Math.round(
-      headerRef.current.scrollLeft /
-        headerRef.current.getBoundingClientRect().width
+      headerRef.current.scrollLeft / headerWidth
     );
 
-    setSelectedTabIndex(currentTabIndex);
+    if (headerWidth < 544) {
```
diogomateus, line 65: 544 is the mobile breakpoint where the table collapses to a single column table.

kimkwanka:
🤞

PR #219, opened 2023-05-24
refactor(auto-suggest-input): Allow to pass input props to AutoSuggestInput

diogomateus:
### What this PR does
Allow to pass input props to AutoSuggestInput.
This can be useful, as an example, to set autocomplete property off.

### How to test?
Pass a valid HTML input prop and see that it's rendered on the input component.

### Checklist:

- [x] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/input/autoSuggestInput/index.tsx
+++ b/src/lib/components/input/autoSuggestInput/index.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import Autosuggest from 'react-autosuggest';
+import Autosuggest, { RenderInputComponentProps } from 'react-autosuggest';
 
 import styles from './style.module.scss';
 import { Option } from '../../../models/autoSuggestInput';
@@ -15,6 +15,7 @@ export default ({
   placeholder,
   className,
   wrapText,
+  inputProps
 }: {
   currentOption: string;
   suggestions: Option[];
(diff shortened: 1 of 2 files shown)


src/lib/components/input/autoSuggestInput/index.tsx
```diff
@@ -47,9 +49,9 @@ export default ({
 
   const getSuggestionValue = (suggestion: Option) => suggestion.value;
 
-  const renderInputComponent = (inputProps: Omit<InputProps, 'ref'>) => (
+  const renderInputComponent = (autoSuggestsInputProps: Omit<InputProps, 'ref'>) => (
```
kimkwanka, line 52:
```suggestion
  const renderInputComponent = (autoSuggestInputProps: Omit<InputProps, 'ref'>) => (
```

src/lib/components/input/autoSuggestInput/index.tsx
```diff
@@ -47,9 +49,9 @@ export default ({
...
     <Input
-      {...inputProps}
+      {...autoSuggestsInputProps}
```
kimkwanka, line 54:
```suggestion
      {...autoSuggestInputProps}
```

kimkwanka (approved):

PR #221, opened 2023-05-24
fix(comparison-table): Fix comparisonTable maxWidth on mobile

diogomateus:
### What this PR does
Fix comparisonTable maxWidth on mobile, which doesn't trigger on mobile devices. This seems to be only visible on website, not on storybook.

### How to test?
`Spotted this issue on public LP on mobile. The name appears after scrolling horizontally a bit `
File

### Checklist:

- [x] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/comparisonTable/components/Row/style.module.scss
+++ b/src/lib/components/comparisonTable/components/Row/style.module.scss
@@ -8,7 +8,7 @@ $cell-min-width: var(--cellWidth, 256px);
 
   & > div {
     width: var(--tableWidth);
-    max-width: var(--tableWidth);
+    max-width: 100vw;
 
     &:nth-child(n + 3) {
       margin: 0;
@@ -16,6 +16,7 @@ $cell-min-width: var(--cellWidth, 256px);
 
     @include p-size-tablet {
       width: $cell-min-width;
+      max-width: var(--tableWidth);
       scroll-snap-align: unset;
     }
   }


kimkwanka (changes requested):
Isn't this the exact change that was [done here](https://github.com/getPopsure/dirty-swan/pull/186/files) and then [got reverted](https://github.com/getPopsure/dirty-swan/pull/202) because it borked something else?

diogomateus:
> Isn't this the exact change that was [done here](https://github.com/getPopsure/dirty-swan/pull/186/files) and then [got reverted](https://github.com/getPopsure/dirty-swan/pull/202) because it borked something else?

I didn't know about this change from @busypeoples, only that this was changed to try to fix the table flaky styles.
@busypeoples do you remember this fix and what it was breaking?

busypeoples:
> > Isn't this the exact change that was [done here](https://github.com/getPopsure/dirty-swan/pull/186/files) and then [got reverted](https://github.com/getPopsure/dirty-swan/pull/202) because it borked something else?
> 
> I didn't know about this change from @busypeoples, only that this was changed to try to fix the table flaky styles. @busypeoples do you remember this fix and what it was breaking?

@diogomateus Can't recall in detail why the change wasn't working, but from what I recall is that the comparison table looked broken on mobile.

diogomateus:
@kimkwanka @busypeoples 

A bit of context to understand this update.

1. The current state is that on mobile we have an issue where we need to swipe to get the correct header values as Andrey reporterd:
Screenshot 2023-05-30 at 10 11 58

2. The revert that Ali did was to fix this issue with aligning no tabled and slightly on mobile:
Screenshot 2023-05-30 at 10 10 52
Screenshot 2023-05-30 at 10 10 47

3. I tested this linking with `website` and my solution is to use 100vw but only on mobile as when it's on mobile the header is full width (there's already a 16px padding). This ensures that the only thing changing is the mobile, as everything else seems to be working as expected.
Screenshot 2023-05-30 at 10 10 29

kimkwanka (approved):

PR #237, opened 2023-06-20
chore(icons): Add more icons

busypeoples:
### What this PR does

Add more icons


busypeoples:
> @busypeoples I'm approving but be aware that this file/folder structure will be changed soon to accommodate our needs. Is it easy to keep track of changes where you're adding the images?

Thanks @diogomateus - there was no problem yet. The more important part is that if the image constants change, there will be some work needed, but that should also be no issue.

PR #239, opened 2023-06-30
docs(input): Add storybook controls to Input stories

diogomateus:
### What this PR does
This PR adds storybook controls to Input stories.
http://localhost:9009/?path=/docs/jsx-inputs-input--input-story

Screenshot 2023-06-30 at 08 37 56

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge


src/lib/components/input/stories/config.ts
```diff
@@ -0,0 +1,56 @@
...
+    defaultValue: false,
+    control: { type: 'boolean' }
+  },
+  prefix: {
+    description: 'A preffix that can be displayed on the left of the input. Specially useful for currencies.',
+    defaultValue: '£',
+    control: { type: 'text' }
+  },
+  id: {
+    description: 'Unique ID identifier of the input. Relevent for accessibility.',
```
talo242, line 33: I think you have a typo here
```suggestion
    description: 'Unique ID identifier of the input. Relevant for accessibility.',
```

src/lib/components/input/input.stories.tsx
```diff
@@ -0,0 +1,135 @@
...
+      hideLabel={hideLabel}
+      prefix={prefix}
+      error={error}
+    />
+  );
+};
+
+InputStory.storyName = 'Input';
+
+export const InputWithPlaceholderAndNoLabel = ({
```
talo242, line 48: After playing a bit with the story, I'm wondering if we really need these different stories.
In my opinion is more useful to see how the component changes using the controls. In this PR, the controls only work with the default Input and this could be confusing (for example if I click on the "Input with prefix" story I cannot edit the prefix using the controls)



What do you think?

src/lib/components/input/stories/config.ts
```diff
@@ -0,0 +1,56 @@
...
+    defaultValue: 'Placeholder',
+    control: { type: 'text' }
+  },
+  label: {
+    description: 'Label of the Input component',
+    defaultValue: 'Input Label',
+    control: { type: 'text' }
+  },
+  hideLabel: {
+    description: 'Wether a label should be hidden. This is needed for accessibility purposes and a label should always be provided',
```
talo242, line 23:
```suggestion
    description: 'Whether or not a label should be hidden. This is needed for accessibility purposes and a label should always be provided',
```

src/lib/components/input/stories/config.ts
```diff
@@ -0,0 +1,56 @@
...
+    defaultValue: false,
+    control: { type: 'boolean' }
+  },
+  prefix: {
+    description: 'A preffix that can be displayed on the left of the input. Specially useful for currencies.',
```
talo242, line 28:
```suggestion
    description: 'A prefix that can be displayed on the left of the input. Specially useful for currencies.',
```

talo242 (changes requested):

src/lib/components/input/input.stories.tsx
diogomateus: Yeah, that sounds good! I usually just use stories to show different examples that are worth showing (I guess it's more useful for stuff like ComparisonTable where it's more complex to show in a single place)

I'll remove these stories, there's not much value added here

src/lib/components/input/stories/config.ts
diogomateus: Thanks for catching all those typos 😅 😅

kimkwanka: That one works great for me: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker
You wouldn't believe how much stuff it uncovers haha

diogomateus: Ohh I'll get this one, I think it's relevent to me 😄

talo242 (approved):

PR #256, opened 2023-08-23
fix(comparison-table): Fix comparison table not setting ref on load

diogomateus:
### What this PR does
This PR fixes comparison table not setting ref on load.

This is a **temporary solution as react-scroll-sync isn't maintained anymore**. 
There is an active issue in the repo that fixes our needs https://github.com/okonet/react-scroll-sync/issues/79 and we should be able to use ref and not IDs once this is released.

### How to test?
Go to a comparison table component on mobile, hard refresh and click on the right arrow (it will only work after you swipe on the content). 
This happens because onSync only triggers after scrolling changes and not on load and the way react-scroll-sync works, we cannot use refs for internal divs 😭 

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/comparisonTable/hooks/useComparisonTable.ts
+++ b/src/lib/components/comparisonTable/hooks/useComparisonTable.ts
@@ -3,6 +3,8 @@ import { useCallback, useEffect, useRef, useState } from 'react';
 
 import { ArrowValues } from '../components/TableArrows';
 
+const headerId = 'comparison-table-header';
+
 export const useComparisonTable = () => {
   const [showMore, setShowMore] = useState<boolean>(false);
   const [headerWidth, setHeaderWidth] = useState(1400);
@@ -110,8 +112,21 @@ export const useComparisonTable = () => {
     setShowMore(!showMore);
   };
 
+  useEffect(() => {
+    if (headerRef.current) {
+      return;
+    }
+
+    const headerById = document.getElementById(headerId);
+console.log('headerById :>> ', headerById);
+    if (headerById) {
+      scrollContainerCallbackRef(headerById);
+    }
+  }, [scrollContainerCallbackRef]);
+
   return {
     headerWidth,
+    headerId,
     contentContainerRef,
     selectedSection,
     setSelectedSection,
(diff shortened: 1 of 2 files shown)


src/lib/components/comparisonTable/hooks/useComparisonTable.ts
```diff
@@ -110,8 +112,21 @@ export const useComparisonTable = () => {
     setShowMore(!showMore);
   };
 
+  useEffect(() => {
+    if (headerRef.current) {
+      return;
+    }
+
+    const headerById = document.getElementById(headerId);
+console.log('headerById :>> ', headerById);
```
kimkwanka, line 121: Leftover console.log :)

src/lib/components/comparisonTable/hooks/useComparisonTable.ts
```diff
@@ -110,8 +112,21 @@ export const useComparisonTable = () => {
     setShowMore(!showMore);
   };
 
+  useEffect(() => {
+    if (headerRef.current) {
+      return;
+    }
+
+    const headerById = document.getElementById(headerId);
```
kimkwanka, line 120: The question is, what will happen if you use two of them? The headerId would be the same in that case, right?

src/lib/components/comparisonTable/hooks/useComparisonTable.ts
diogomateus: Linter needs some improvement 😅

src/lib/components/comparisonTable/hooks/useComparisonTable.ts
diogomateus: Ah, true, great catch! 
I started by using `useId` but our react version doesn't have that yet. I'll use our own `generateId` function here

diogomateus: Fixed 🔨

kimkwanka (approved):

talo242 (approved):
Tested locally and works as expected 👍🏼

PR #259, opened 2023-09-04
fix(multidropzone): [EMU-5919] add label to multidropzone component

a-maffei:
### What this PR does

1. Add a `label` associated to the `input` to upload a file
2. Update the label style (display) to maintain current layout
3. Add a test to check that the label text and the input are connected (which in a second stage could be used to refactor the tests that use `data-testid`)

### Why is this needed?

To make the component more testable (as mentioned above) and accessible. 

Solves:
[EMU-5919](https://linear.app/feather-insurance/issue/EMU-5919/add-label-to-multidropzone-component)

### How to test?

You can run the test with the title "Copy text should associate input with its label".

### Checklist:

- [x] I reviewed my own code

### Browser support

- [ ] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/multiDropzone/index.tsx
+++ b/src/lib/components/multiDropzone/index.tsx
@@ -113,7 +113,9 @@ const MultiDropzone = ({
         />
         <label
           htmlFor={uniqueId.current}
-          className={`p-h4 mt8 d-block ${isCondensed ? styles.textInline : ''}`}
+          className={`p-h4 mt8 d-block c-pointer ${
+            isCondensed ? styles.textInline : ''
+          }`}
         >
           {uploading
             ? textOverrides?.currentlyUploadingText ||
--- a/src/lib/components/multiDropzone/index.tsx
+++ b/src/lib/components/multiDropzone/index.tsx
@@ -88,6 +88,8 @@ const MultiDropzone = ({
     onDrop,
   });
 
+  const [uniqueId] = useState(generateId());
+
   return (
     <div className={styles.container}>
       <div
@@ -99,18 +101,14 @@ const MultiDropzone = ({
         )}
         {...getRootProps()}
       >
-        <input
-          data-testid="ds-drop-input"
-          id="ds-drop-input"
-          {...getInputProps()}
-        />
+        <input data-testid="ds-drop-input" id={uniqueId} {...getInputProps()} />
         <img
           className={isCondensed ? styles.img : ''}
           src={isCondensed ? icons.uploadSmallIcon : icons.uploadIcon}
           alt="purple cloud with an arrow"
         />
         <label
-          htmlFor="ds-drop-input"
+          htmlFor={uniqueId}
           className={`p-h4 mt8 d-block ${isCondensed ? styles.textInline : ''}`}
         >
           {uploading
--- a/src/lib/components/multiDropzone/index.tsx
+++ b/src/lib/components/multiDropzone/index.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState } from 'react';
+import { useCallback, useState, useRef } from 'react';
 import classnames from 'classnames';
 import { useDropzone, FileRejection } from 'react-dropzone';
 import AnimateHeight from 'react-animate-height';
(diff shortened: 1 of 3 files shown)


src/lib/components/multiDropzone/index.tsx
```diff
@@ -100,38 +101,45 @@ const placeholder = `${textOverrides?.supportsTextShort || "Supports"} ${fileLis
       >
         <input
           data-testid="ds-drop-input"
+          id="ds-drop-input"
```
diogomateus, line 104: This should be code-generated because there might be more than one instance of the input and HTML Ids should be unique.

You can use our own generateId function:
```
    const [uniqueId] = useState(generateId());
```

diogomateus (changes requested):
No problem with prettier!

src/lib/components/multiDropzone/index.tsx
a-maffei: Of course, makes complete sense that it should be unique. I updated it as you suggested. For now, I left  `data-testid` - once everything works with the label, I can update the tests so they use the label text instead (if it's part of the scope of this ticket).

talo242: @diogomateus do we need to store the `generateId()` in a `useState` hook?
If it's to prevent this ui to get re-generated on every render, I'm wondering if we can store the variable outside of the component function.
E.g.
```tsx
const multiDropzoneId = generateId();

const MultiDropzone = (props) => {
  ...
  <input
      id={multiDropzoneId}
}
```

diogomateus: You're right, the reason is to prevent from re-generating on every render.
Wouldn't storing it outside the component make it always the same for every MultiDropzone import happening?

We could also use `useRef` as it doesn't change unless it's specified.

diogomateus: Just a side note: when we upgrade react we could use the [useId function from react ](https://react.dev/reference/react/useId)

talo242: Ohhh yes, I didn't think about that 😅. Using a `ref` makes more sense.

src/lib/components/multiDropzone/index.tsx
```diff
@@ -73,20 +75,21 @@ const placeholder = `${textOverrides?.supportsTextShort || "Supports"} ${fileLis
...
 
-
   const { getRootProps, getInputProps } = useDropzone({
     accept: formattedAccept,
     disabled: uploading,
     maxSize,
     onDrop,
   });
 
+  const [uniqueId] = useState(generateId());
```
diogomateus, line 91: Can you update to use the code below?
You can check the discussion on the other thread of this PR
```suggestion
  const uniqueId = useRef(generateId());
```

a-maffei: Thank you @diogomateus, @busypeoples, @talo242 - I've updated the method to store a unique id for every import of the component and preserve the value across rerenders with `useRef`.

talo242 (approved):

busypeoples (approved):
Approved, this all looks very sensible.
The only thing I noticed is that the label has a default cursor not a pointer cursor when we hover or click on the label. This is the only thing I noticed comparing the before and after implementation. Just double checking if this is the expected behaviour.

a-maffei:
> Approved, this all looks very sensible. The only thing I noticed is that the label has a default cursor not a pointer cursor when we hover or click on the label. This is the only thing I noticed comparing the before and after implementation. Just double-checking if this is the expected behaviour.

Thanks for noticing @busypeoples! This wasn't the expected behavior: you should see a pointer cursor as you hover over any part of the component. The `label` is a child of a `div` with the correct cursor type (pointer), which isn't overriding the default one. 

I updated the code which now works as expected and asked you for another review.

busypeoples (approved):

busypeoples:
> > Approved, this all looks very sensible. The only thing I noticed is that the label has a default cursor not a pointer cursor when we hover or click on the label. This is the only thing I noticed comparing the before and after implementation. Just double-checking if this is the expected behaviour.
> 
> Thanks for noticing @busypeoples! This wasn't the expected behavior: you should see a pointer cursor as you hover over any part of the component. The `label` is a child of a `div` with the correct cursor type (pointer), which isn't overriding the default one.
> 
> I updated the code which now works as expected and asked you for another review.

Thanks, approved.

PR #262, opened 2023-09-06
fix(address-autocomplete): [EMU-5932] make “enter address manually" accessible

a-maffei:
### What this PR does

1. Replaces the `<span>` element with a `<button>`
2. Adds additional styling to the button for it to align with the original design
3. Updates to lowercase the first letter of the button copy in English - both in JSX and in the test file (in German it should stay uppercase)

EMU-5932-eng
EMU-5932-deu

### Why is this needed?

Solves:
[EMU-5932](https://linear.app/feather-insurance/issue/EMU-5932/address-autocomplete-component-make-enter-address-manually-accessible)

### How to test?

In the JSX/AutocompleteAddress component.

### Checklist:

- [x] I reviewed my own code
- [x]The changes align with the designs I received  

### Browser support

- [ ] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/autocompleteAddress/index.tsx
+++ b/src/lib/components/autocompleteAddress/index.tsx
@@ -343,7 +343,7 @@ const AutocompleteAddress = ({
         <div className="p-p mt8">
           {manualAddressEntryTexts?.preText || 'Or '}
           <button
-            className={classNames(`p-a p-p fw-bold c-pointer bg-transparent`)}
+            className={'p-a p-p fw-bold c-pointer bg-transparent'}
             onClick={handleEnterAddressManually}
             type="button"
           >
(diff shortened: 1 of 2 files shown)


src/lib/components/autocompleteAddress/index.tsx
```diff
@@ -339,12 +342,13 @@ const AutocompleteAddress = ({
       {manualAddressEntry === false && (
         <div className="p-p mt8">
           {manualAddressEntryTexts?.preText || 'Or '}
-          <span
-            className="p-a fw-bold c-pointer"
+          <button
+            className={classNames(`p-a p-p fw-bold c-pointer bg-transparent`)}
```
mbaguiar, line 346: Just a little nitpick, we don't need the `classNames` package or the template string here, we can leave it as a normal string

mbaguiar (approved):

src/lib/components/autocompleteAddress/index.tsx
a-maffei: Makes perfect sense, I'll make this tiny tweak (without asking you for a new review). Thank you!

talo242 (approved):
💅🏼

PR #266, opened 2023-09-15
fix(address): prevent infinite loop when initial address is not provided

talo242:
### What this PR does

1. This PR removes the `onAddressChange` callback from the `useEffect` to prevent an infinite loop when `address` prop is not provided.
2. It adds the `name` attribute to the inputs to better improve a11y
3. It provides an `inputProps` prop in order to modify the `name` attribute and any other future values we seem necessary in the future
4. It also reformats the AutocompleteAddress story

Solves:
EMU-6719

### How to test?
1. To test this, you have to modify the `AutocompleteAddressStory` function and remove the address prop:
```tsx
  return (
    <AutocompleteAddress
      // address={address}
      apiKey={apiKey}
      manualAddressEntryTexts={manualAddressEntryTexts}
      onAddressChange={handleOnAddressChange}
      placeholders={placeholders}
    />
  );
```
2. Go go the storypage on Storybook and enter an address
6. The autocomplete should work and no infinite loop should be triggered.
If you try this same steps on the `main` branch you'd get an infinite loop.

### Checklist:

- [x] I reviewed my own code
No visual changes
- [ ] I have attached screenshot(s), video(s) or gif(s) showing that the solution is No visual changes
No new media

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari

### DIFF
--- a/src/lib/components/autocompleteAddress/index.stories.tsx
+++ b/src/lib/components/autocompleteAddress/index.stories.tsx
@@ -9,49 +9,55 @@ const story = {
     address: {
       description: 'The address properties',
       table: {
-        type: { 
-          summary: 'Partial<Address>'
+        type: {
+          summary: 'Partial<Address>',
         },
       },
     },
     apiKey: {
       defaultValue: 'AIzaSyDg0DSrjYKt5smmsjkVasDz7c4T5rbOXT8',
-      description: 'Your private API key for the [Google Places API](https://developers.google.com/maps/documentation/places/web-service/overview)',
+      description:
+        'Your private API key for the [Google Places API](https://developers.google.com/maps/documentation/places/web-service/overview)',
       table: {
-        type: { 
-          summary: 'Partial<Address>'
+        type: {
+          summary: 'Partial<Address>',
         },
       },
     },
     onAddressChange: {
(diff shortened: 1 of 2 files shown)


kimkwanka (approved):

talo242:
> Tested it locally, both the prior and this version with the `address={address}` line commented out, but couldn't replicate an infinite loop. (Modified the AutocompleteAddress story, as instructed)
> 
> image
> 
> So not sure what's up with that 😕

Mmm weird 🤔 I'm pretty sure the infinite loop came from this, as I was able to reproduce it multiple times.
I'm gonna merge this and test it on the Referral payout page (that was the page where I first spotted the bug)

Thanks for your review @kimkwanka!

kimkwanka:
> > Tested it locally, both the prior and this version with the `address={address}` line commented out, but couldn't replicate an infinite loop. (Modified the AutocompleteAddress story, as instructed)
> > image
> > So not sure what's up with that 😕
> 
> Mmm weird 🤔 I'm pretty sure the infinite loop came from this, as I was able to reproduce it multiple times. I'm gonna merge this and test it on the Referral payout page (that was the page where I first spotted the bug)
> 
> Thanks for your review @kimkwanka!

Yeah, not sure what's going on here 🤷‍♀️

PR #271, opened 2023-10-17
refactor(icon): [STO-6158] replace internal icons with Icon component

a-maffei:
### What this PR does

In the DS components that use "internal icons" (not icons passed as props or used with external libraries), we replace the current solution with the up-to-date `Icon` component. 

The components are:

- CardButton
- Chip (only close button, not left icon)
- Date Selector (only date icon)
- Download button (only the checkmark icon, not the download icon as this is passed as prop to the Button component)
- InformationBox
- Modal (all types)
- Multidropzone

The design doesn't match 100% for some components as default sizes and icon shapes don't always overlap with the original.

**CardButton** 
Before
CardButton-b

After
CardButton-a

**Chip** 
Before
Chip-b

After
Chip-a

**DateSelector** 
Before
DateSelector-b

After
DateSelector-a

**DownloadButton** 
Before
Download-b

After
Download-a

**InformationBox** 
Before
InformationBox-a

After
InformationBox-b

**Modal** 
Before
RegularModal-b

After
RegularModal-a

**Multidropzone** 
Before
Multidropzone-1-b

After
Multidropzone-1-a

### Why is this needed?

Solves:
[STO-6158](https://linear.app/feather-insurance/issue/STO-6158/clean-dirty-swan-components-to-use-the-up-to-date-icon-component)

### How to test?
- Start Storybook
- Go by one by one through the stories for the components listed above
- Compare them with the version in production

### Checklist:

- [x] I reviewed my own code

### Browser support

- [ ] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Edge


src/lib/components/chip/index.tsx
```diff
@@ -25,16 +25,14 @@ export default ({
...
-        src={removeButtonHighlightedIcon}
-        alt="removal x button highlighted"
-      />
-      <img
-        className={styles['chip-remove-button']}
-        src={removeButtonIcon}
-        alt="removal x button"
-      />
+      <XIcon color={'purple-300'} className={styles['chip-remove-button']} />
+      {
```
diogomateus, line 29: I think this `{` is not needed.

I think this could also be changed to have just on `XIcon` and remove the `chip-remove-button-highlighted` className. Something like:

```
 <div
      className={`c-pointer ${styles['chip-button-container']}`}
      onClick={() => onRemove(value)}
    >
      <XIcon className={styles['chip-remove-button']}  />
      </div>
```

```
.chip-button-container {
  color: $ds-purple-300;
  
  &:hover {
    color: $ds-purple-500;
    }
}
```

diogomateus: Also check this XIcon, it seems to not be vertically centred anymore

src/lib/components/dateSelector/index.tsx
```diff
@@ -238,12 +238,17 @@ export const DateSelector = ({
           className={styles['date-calendar-container']}
           ref={calendarContainerRef}
         >
-          <img
-            className="c-pointer"
-            src={calendarIcon}
-            alt="calendar"
+          <button
+            type="button"
```
diogomateus, line 242: 🙌 🙌 🙌

src/lib/components/informationBox/index.tsx
```diff
@@ -19,47 +22,55 @@ const InformationBox = ({
...
+        styles[`informationBox--${variant}`]
+      )}
+      role="alert"
+    >
+      <div className="d-flex">
+        {showIcon && (
+          <div data-testid="information-box-icon" className={'mr8'}>
+            <AlertCircleIcon
+              className={styles.icon}
+              color={mapAlertIconColor[variant]}
```
diogomateus, line 50: I think this was wrong before, can you add size 24 if props.size is default and 16 when small?

diogomateus (commented):
Just a couple of suggestions, but looking good! 🙌

src/lib/components/chip/index.tsx
a-maffei: @diogomateus That makes sense, and you are right: the `XIcon` was falling below the center. It took me a while to fix it, and it only worked when I got to styling the SVG itself. Not sure if there's a simpler fix I missed. Anyway, I worked on the changes you suggested both here and in `informationBox`!

src/lib/components/multiDropzone/UploadFileCell/index.tsx
```diff
@@ -2,7 +2,8 @@ import React from 'react';
 import classnames from 'classnames';
 
 import styles from './style.module.scss';
-import icons from '../icons/index';
```
diogomateus, line 5: We can now remove these icons, right? The whole folder 👀

a-maffei: Yes, thanks for catching it, we can delete them all - done ✅

PR #272, opened 2023-10-18
feat(toggle): Add Toggle component

diogomateus:
### What this PR does
This PR adds the Toggle component.

### How to test?
Run `yarn storybook` and check [toggle component story](http://localhost:9009/?path=/docs/jsx-inputs-toggle--toggle-story).

**Preview:**
Screenshot 2023-10-18 at 16 02 51
Screenshot 2023-10-18 at 16 02 58

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/input/toggle/index.stories.tsx
+++ b/src/lib/components/input/toggle/index.stories.tsx
@@ -1,0 +1,191 @@
+
+import { useState } from 'react';
+import { Toggle, ToggleProps } from '.';
+
+const story = {
+  title: 'JSX/Inputs/Toggle',
+  component: Toggle,
+  argTypes: {
+    options: {
+      description: 'Object that contains the possible options for rendering in the input.',
+      defaultValue: {
+        CAT:{
+          title: 'Cat',
+          description: 'At least 1'
+        },
+        DOG:{
+          title: 'Dog',
+          description: 'At least 2'
+        },
+        NONE:{
+          title: 'None',
+          description: 'No pets'
+        }
+      }
+    },
+    value: {
+      description: 'Current checked values.',
+    },
+    onChange: {
+      description: 'Function called everytime a value changes.',
+      action: true,
+      table: {
+        category: "Callbacks",
+      },
+    },
+    bordered: {
+      control: { type: 'boolean' },
+      description: 'Property that defines if checkbox should show the border around each label',
+      defaultValue: true
+    },
+    inlineLayout: {
+      description: 'Property that defines if options should show inline instead of block. Check inline checkbox options story for examples.',
+      defaultValue: false
+    },
+    className: {
+      description: 'ClassNames for custom styling',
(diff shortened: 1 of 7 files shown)


diogomateus:
@talo242 I created this using a similar approach to Checkbox/Radio, wdyt?

src/lib/components/input/toggle/index.stories.tsx
```diff
@@ -0,0 +1,191 @@
+
+import { useState } from 'react';
+import { Toggle, ToggleProps } from '.';
+
+const story = {
+  title: 'JSX/Inputs/Toggle',
+  component: Toggle,
+  argTypes: {
+    options: {
+      description: 'Object that contains the possible options for rendering in the input.',
```
a-maffei, line 10: Since we have a story where the toggle has no text, we could add a sentence like "To create an icon-only toggle, pass an empty string" to clarify what to do in this case (though I imagine it won't be common).

src/lib/components/input/toggle/index.tsx
```diff
@@ -0,0 +1,118 @@
+import classNames from "classnames";
+
+import styles from './styles.module.scss';
+export interface ToggleWithDescription {
+  title: string;
+  description?: string;
+}
+
+export interface ToggleProps<ValueType extends string> {
+  options: Record<ValueType, string | ToggleWithDescription>;
```
a-maffei, line 10: Should we update `string` to `ReactNode` to align it with the checkbox/radio buttons or not for now? It might create more confusion - I associate toggles with switching between boolean values but we're not meant to use them here.

a-maffei (approved):
Looks great! I've left a couple of minor questions - one regarding types based on our conversation yesterday.

src/lib/components/input/toggle/index.tsx
diogomateus: I was thinking about doing this after your work but I think I can copy your code here and just be done with this 😄

src/lib/components/input/toggle/index.stories.tsx
```diff
@@ -0,0 +1,191 @@
...
+    bordered: {
+      control: { type: 'boolean' },
+      description: 'Property that defines if checkbox should show the border around each label',
+      defaultValue: true
+    },
+    inlineLayout: {
+      description: 'Property that defines if options should show inline instead of block. Check inline checkbox options story for examples.',
+      defaultValue: false
+    },
+    className: {
```
talo242, line 45: Shouldn't this be `classNames` to match the prop?

```suggestion
    classNames: {
```

talo242 (commented):
@diogomateus sorry for not answering your comment 🙈 

The API looks nice ✨
I just left a tiny comment regarding the documentation.

PR #273, opened 2023-10-18
refactor(input): [STO-6165] update type for checkbox and radio options

a-maffei:
### What this PR does

- Updates the type of the `options` prop used by the Radio and Checkbox component to render the input label, so that it doesn't only accept strings but the broader `ReactNode`
- Adds a test per component to check the label is rendered when we pass it in HTML format

### Why is this needed?

Solves:
[STO-6165](https://linear.app/feather-insurance/issue/STO-6165/update-title-type-for-ds-radio-and-checkbox-input-components)

### How to test?

Run storybook and go through the radio and checkbox component stories to check they still render correctly.

### Checklist:

- [x] I reviewed my own code

### Browser support

- [ ] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/input/checkbox/index.test.tsx
+++ b/src/lib/components/input/checkbox/index.test.tsx
@@ -4,7 +4,10 @@ import { Checkbox, CheckboxProps } from '.';
 
 const mockOnChange = jest.fn();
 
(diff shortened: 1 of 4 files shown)


diogomateus (approved):
🙌

PR #276, opened 2023-10-20
feat(button): Refactor button component

diogomateus:
### What this PR does
This PR refactors the button component to use up to date icons, new variant names and replacing props names for a more standardised approach (children vs buttonTitle allowing ReactNodes).

### How to test?
Run storybook and [run button story](http://localhost:9009/?path=/docs/jsx-button--button-story).

**Preview:**
Screenshot 2023-10-20 at 10 21 40
Screenshot 2023-10-20 at 10 21 35
Screenshot 2023-10-20 at 10 21 45

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge


kimkwanka (approved):

src/lib/models/styles.ts
```diff
@@ -56,4 +56,4 @@ export type Color = 'transparent' | 'blue-100'
 | 'yellow-700'
 | 'yellow-900';
 
-export type Size = 0 | 8 | 16 | 24 | 32 | 40 | 48 | 56 | 64 | 72 | 80 | 88 | 96;
\ No newline at end of file
+export type Size = 0 | 8 | 16 | 20 | 24 | 32 | 40 | 48 | 56 | 64 | 72 | 80 | 88 | 96;
```
ruizhen88, line 59: just curious why we are breaking 8x scales rule. was it a design decision?

diogomateus: I'm syncing with Andrey today and will check why this is happening, I'll post the feedback here to keep us informed

src/lib/components/downloadButton/index.tsx
```diff
@@ -14,20 +13,22 @@ export interface DownloadButtonProps {
 const InitialButton = ({ onDownload }: { onDownload: () => void }) => (
   <Button
     className={`w100 ${styles.button}`}
-    buttonTitle="Download"
-    leftIcon={{ src: downloadIcon, alt: 'download arrow icon' }}
+    leftIcon={<DownloadIcon />}
```
talo242, line 16: I'm still wondering if having specific `leftIcon` and `rightIcon` props are necessary?

If we are allowing `children` wouldn't it be enough to pass the icons there?

E.g.
```tsx
 <Button
    className={`w100 ${styles.button}`}
    onClick={onDownload}
    data-cy="download-documents-button"
  >
  <DownloadIcon className="mr8"/> Download
</Button>
```
Or something similar?

It makes sense for input components, since adding stuff inside the input is very annoying, but with buttons there is not much value being added. Also moving towards this direction makes the component less flexible.

What do you think?

talo242 (approved):
Just a comment about deprecating `leftIcon` prop, but that change could be introduced in a subsequent PR so this one is good to go :+1:

src/lib/models/styles.ts
talo242: I've seen that design is using more and more smaller units. E.g. `4px` or `12px`.

For me, it makes sense to have finer control on smaller amounts, to let designers have more flexibility tweaking small pieces of UI. So I don't see a problem with adding `20`.

I'd go as far as adding `4` and `12` to the list of allowed sizes and share with Design that this is the list we should stick to.
```suggestion
export type Size = 0 | 4 | 8 | 12 | 16 | 20 | 24 | 32 | 40 | 48 | 56 | 64 | 72 | 80 | 88 | 96;
```

ruizhen88: totally agreed, i don't have problem breaking 8x either. i think we could even ask designers for a list of common scales, align our code to the list and ask them to stick with it.

src/lib/components/downloadButton/index.tsx
diogomateus: For me it's about having a consistent design for buttons without having to worry about it when implementing buttons.

Two examples I can think of are: 
How can you make sure that Icons always have the same size and spacing between icons and text?
Or if we have a `8` spacing and we want to switch to 16 in the future, we need to go through every component usage to make it have a standard logic, which IMO, should be contained within the button.

Also `children` is still accepting `ReactNode` and still allows you to add anything you want in there. WDYT?

a-maffei:
While reviewing the new Card component, I observed that the button text overflows vertically on smaller screens, especially when the copy is lengthy. A solution could be setting the button's height to `auto` and adjusting its padding appropriately. The rest looks great!

Screenshot 2023-10-25 at 13 24 17

diogomateus:
> While reviewing the new Card component, I observed that the button text overflows vertically on smaller screens, especially when the copy is lengthy. A solution could be setting the button's height to `auto` and adjusting its padding. The rest looks great!
> 
> Screenshot 2023-10-25 at 13 24 17

I think this should actually be fixed now, at least according to the [visual regression report on app](https://github.com/getPopsure/app/pull/2431)! 🤞 

I think adapting the mobile layout would also help 😅 but I usually don't care that much about it on storybook

src/lib/models/styles.ts
talo242: Maybe it's time to start exploring design tokens, so we are totally aligned between Figma <-> code 👀

src/lib/components/downloadButton/index.tsx
talo242: That's a good point 🤔 
I can think of a scenario when we need an icon with a spacing other than `8`, but I guess that shouldn't happen that often, and in those cases we can fall back to the children.

I'm going to resolve this conversation to keep the PR clean and tidy.
Thanks for your insights!

a-maffei (approved):

src/lib/models/styles.ts
diogomateus: Talked to Andrey and he also agrees to breaking into 4 in the smaller cases (4, 12, 20 and maybe 28).
For now this will only apply to icons but eventually allow us to margin, padding and so on too (once we get something like [PurgeCSS](https://purgecss.com/) working)

diogomateus: And yes to design tokens from figma!!!! 🎉

PR #279, opened 2023-10-24
feat(card): Add Card component

diogomateus:
### What this PR does
This PR adds the new Card component.

### How to test?
Run storybook and go to [Card story](http://localhost:9009/?path=/docs/jsx-card--card-story).

**Preview:**
Screenshot 2023-10-24 at 17 02 30
Screenshot 2023-10-24 at 17 02 37
Screenshot 2023-10-24 at 17 02 40
Screenshot 2023-10-24 at 17 02 43

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/card/index.tsx
+++ b/src/lib/components/card/index.tsx
@@ -1,0 +1,147 @@
+import { ReactNode } from 'react';
+import classNamesUtil from 'classnames';
+import { ChevronRightIcon } from '../icon';
+
+import styles from './style.module.scss';
+
+export interface CardProps {
+  children?: ReactNode;
+  classNames?: {
+    wrapper?: string;
+    label?: string;
+    title?: string;
+    description?: string;
+    children?: string;
+    icon?: string;
+    actionIcon?: string;
+  };
+  density?: 'balanced' | 'compact' | 'spacious';
+  dropShadow?: boolean;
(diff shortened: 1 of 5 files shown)


src/lib/components/card/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+        className={classNamesUtil('c-pointer d-flex w100 br8', styles.button)}
+        onClick={onClick} 
+        type="button"
+      >
+        {renderContent(props)}
+      </button>
+    )
+  }
+
+  return renderContent(props);
```
kimkwanka, line 144: This is technically just a component, so why not use it as such? This can lead to some very hard to debug issues where hooks etc. behave erratic which is why I would advise against using the component as a function.
(I know that in this case it might not make a difference but for consistency I would prefer usage as a component)

diogomateus: Yeah, you're right, that makes way more sense! Just updated 🙌

kimkwanka (approved):

talo242 (commented):
🗃️

talo242 (approved):

a-maffei (approved):

PR #286, opened 2023-12-13
fix(multi-dropzone): [EMU-7000] Fix MultiDropzone cell icon shrinking

diogomateus:
### What this PR does
**Before:**
Screenshot 2023-12-13 at 09 39 50

**After:**
Screenshot 2023-12-13 at 09 39 53

Solves:
EMU-7000

### How to test?

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/multiDropzone/UploadFileCell/index.tsx
+++ b/src/lib/components/multiDropzone/UploadFileCell/index.tsx
@@ -52,11 +52,12 @@ const UploadFileCell: React.FC<Props> = ({
       })}
     >
       <div className={`w100 ${styles['cell-left-section']}`}>
-        <FileIcon
-          className={classnames(`${styles['main-icon']} ${styles.icon}`)}
-          color={mapFileIconColor[uploadStatus]}
-          size={24}
-        />
+        <div className={classnames(`${styles['main-icon']} ${styles.icon}`)}>
+          <FileIcon
+            color={mapFileIconColor[uploadStatus]}
+            size={24}
+          />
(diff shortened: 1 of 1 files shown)


talo242:
I don't think this problem is specific to the `Multidropzone` component, I think it affects all icons. You can see [inside the new PolicyDocuments component](https://github.com/getPopsure/app/blob/7d34dba4224612b0e231b37d4252c4d17a8cdd2d/src/features/policies/components/detail/documentCell/index.tsx#L7) that the `FileTextIcon` is being imported straight from DS.

IMHO the expected behavior is that if you set an explicit `size` attribute, the icon should remain at that size, without having to set a different minimum size in outer wrappers.

What do you think?

diogomateus:
> I don't think this problem is specific to the `Multidropzone` component, I think it affects all icons. You can see [inside the new PolicyDocuments component](https://github.com/getPopsure/app/blob/7d34dba4224612b0e231b37d4252c4d17a8cdd2d/src/features/policies/components/detail/documentCell/index.tsx#L7) that the `FileTextIcon` is being imported straight from DS.
> 
> IMHO the expected behavior is that if you set an explicit `size` attribute, the icon should remain at that size, without having to set a different minimum size in outer wrappers.
> 
> What do you think?

Oh, I see! I didn't realise this was happening, will take a look! Closing this PR for now 👍

PR #289, opened 2024-01-12
fix(card): [EMU-7042] Add falsy value option for actionIcon and fix actionable card sizes

diogomateus:
### What this PR does
This PR fixes:
- card action action which now allows to use falsy values to not render it (e.g. `<Card actionIcon={null} />`)
- fixes card sizes that were differently if it was an action card, with and without action icon.

Solves:
EMU-7042

**Before:**
Screenshot 2024-01-12 at 11 51 19

**After:**
Screenshot 2024-01-12 at 11 47 17

### How to test?
Run storybook and check the [Card stories](http://localhost:9009/?path=/docs/jsx-card--card-story).

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/card/index.tsx
+++ b/src/lib/components/card/index.tsx
@@ -41,87 +41,91 @@ const CardContent = ({
   actionIcon,
   title,
   titleVariant = 'large',
-}: CardProps) => (
-  <section
-    className={classNamesUtil(
-      'd-flex fd-column jc-center br8 bg-white w100 ta-left',
-      { 'bs-sm': dropShadow },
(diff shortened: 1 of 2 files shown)


a-maffei (approved):
I applied `actionIcon={null}` and compared card sizes with and without the action icon - everything works as described 👍

talo242 (approved):
Thanks for this @diogomateus! 💅🏼

PR #293, opened 2024-02-15
Change input type from string to number

ruheni:
### What this PR does

Update default currency input type

### Why is this needed?

Please include relevant motivation and context of why this PR is necessary, sentry/linear/notion/...

Solves:
EMU-7137

### How to test?

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/input/currency/index.tsx
+++ b/src/lib/components/input/currency/index.tsx
@@ -62,7 +62,7 @@ const CurrencyInput = ({
     <Input
       prefix="€"
       ref={inputRef}
-      type="string"
+      type="number"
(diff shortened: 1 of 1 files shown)


diogomateus (changes requested):
Is this for displaying the numeric keyboard? Could we [use inputMode](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode#numeric) instead?

diogomateus (approved):
✅

PR #301, opened 2024-04-05
feat(card): Add polymorphic as prop support to Card component

diogomateus:
### What this PR does
This PR adds the `as` prop to Card component allowing it to be rendered as other components.

This is specially useful if we want to use `a`, `button` or react-router `Link` component props without having to wrap the component in those elements (which allows to make use of hover, focus and related states).

### How to test?
Run storybook and [go to stories card, to the `Card as other components` story](http://localhost:9009/?path=/docs/jsx-card--card-story#card-as-other-components).

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/card/index.tsx
+++ b/src/lib/components/card/index.tsx
@@ -1,10 +1,14 @@
-import { ReactNode } from 'react';
+import { ElementType, ReactNode } from 'react';
 import classNamesUtil from 'classnames';
 import { ChevronRightIcon } from '../icon';
 
 import styles from './style.module.scss';
 
-export interface CardProps {
+const CardDefault = 'section' as const
+type CardDefaultAsType = typeof CardDefault;
+
+type CardOwnProps<E extends ElementType = CardDefaultAsType> = {
+  as?: E;
   children?: ReactNode;
   classNames?: {
     buttonWrapper?: string;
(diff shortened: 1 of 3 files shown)


src/lib/components/card/index.tsx
```diff
@@ -1,10 +1,14 @@
...
 import { ChevronRightIcon } from '../icon';
 
 import styles from './style.module.scss';
 
-export interface CardProps {
+const CardDefault = 'section' as const
+type CardDefaultAsType = typeof CardDefault;
+
+type CardOwnProps<E extends ElementType = CardDefaultAsType> = {
+  as?: E;
```
ruheni, line 11: Not necessarily a blocker from merging the PR but, would we want to narrow the type of elements we could render the `Card` component as — i.e, buttons, anchor, section, etc — or do we want to keep it wide? 🤔

ruheni (approved):

src/lib/components/card/index.tsx
diogomateus: That's a very good point! My main concern about this is on how to do it on a way that doesn't block react components like `react-router` or `next` Link components, WDYT?

a-maffei (approved):

src/lib/components/card/index.tsx
ruheni: That makes sense. Let's leave it as it is :)

PR #309, opened 2024-05-13
fix(TabInfoButton): [EMU-7334] Increase contrast

kimkwanka:
### What this PR does

This PR increases the contrast ratio between TabInfoButton and background as user testing showed that it's really hard to see.

#### Before

image

#### After

image

### Why is this needed?

Solves:
EMU-7334

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [xw] Edge

### DIFF
--- a/src/lib/components/comparisonTable/components/TableInfoButton/style.module.scss
+++ b/src/lib/components/comparisonTable/components/TableInfoButton/style.module.scss
@@ -25,7 +25,7 @@
   }
 
   svg {
-    fill: $ds-primary-100;
+    fill: $ds-grey-400;
     transition: all 0.3s ease;
   }
 
--- a/src/lib/components/comparisonTable/components/TableInfoButton/style.module.scss
+++ b/src/lib/components/comparisonTable/components/TableInfoButton/style.module.scss
@@ -20,12 +20,16 @@
     background-color: $ds-primary-50;
 
     svg {
-      fill: $ds-primary-300;
+      path {
+        fill: $ds-primary-300;
+      }
     }
   }
 
   svg {
-    fill: $ds-grey-400;
+    path {
+      fill: $ds-grey-400;
+    }
     transition: all 0.3s ease;
   }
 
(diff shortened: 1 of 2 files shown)


diogomateus (approved):

busypeoples (approved):

src/lib/components/comparisonTable/components/TableInfoButton/style.module.scss
```diff
@@ -25,7 +25,7 @@
   }
 
   svg {
-    fill: $ds-primary-100;
+    fill: $ds-grey-400;
```
diogomateus, line 28: Not necessary as part of this PR but if you want, you can also replace this inline SVG with
`<InfoInfo color="grey-400" size={20 /* or other */} />

kimkwanka: Sounds nifty, but how would I then give it hover states and similar? I tried manipulating it with

```
svg {
 fill: SOMECOLOR
}
```

to no avail :/
Looks like the fill is set to "currentColor" at all times

kimkwanka: Nvm, I need to use svg -> path instead.

diogomateus: You can then remove the svg fill color as the Icon component captures the button color (that's what currentColor is doing):
```
.button {
  color: $ds-primary-400;
  
  &:hover {   
    color: $ds-primary-300;
  }
}
```

kimkwanka: Ah cool 👍 Will keep that in mind and use it in the future! Here however, for better a11y, I wrapped the icon in a real button and just tweaked the original custom css instead of creating 2 new classes (one for the wrapping button and one for the icon itself).

PR #313, opened 2024-05-15
refactor(date-selector): Refactor DateSelector component

diogomateus:
### What this PR does
This PR refactors the date selector component to have 3 text inputs instead of 3 select inputs.
By setting 3 inputs it means that keyboard/typing navigation also needs some custom adjustments:

- when reaching the end of input, focus changes to next input:
ezgif-7-a14f96d45c

- when using right/left arrows, it is possible to navigate to next/previous inputs (like tab and shit+tab). if there is a value already, input value is selected
ezgif-7-37e49bad7e

- backspace, if current focused input is empty, goes to previous input:
ezgif-7-c31fadbe62

- if day is 4 or more focus changes to month input, if month is 2 or more focus changes to year input:
ezgif-7-85b713f993

### How to test?
Test visiting the story:
http://localhost:9009/?path=/docs/jsx-dateselector--date-selector-story

**Before:**
Screenshot 2024-05-16 at 08 37 02

**After:**
Screenshot 2024-05-16 at 08 36 53

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge


busypeoples:
Two things I noticed: You could set a `min` of 1 and `max` of 12. Also if the year is invalid, you could highlight only the year as wrong so it is clear where the error is.

Another question: this will break all our end to end tests?

kimkwanka (approved):

src/lib/components/dateSelector/components/Calendar.tsx
```diff
@@ -0,0 +1,112 @@
...
+          showOutsideDays={true}
+          fromMonth={dateCalendarFromMonth}
+          toMonth={dateCalendarToMonth}
+          selectedDays={selectedDateInDateType}
+          onDayClick={(date: Date) => {
+            if (
+              dayjs(date).isAfter(dateCalendarFromMonth) ||
+              dayjs(date).isBefore(dateCalendarToMonth)
+            ) {
+              const selectedDate = dayjs(date).format(dateFormat);
```
talo242, line 93: It seems like we are not checking if `dateFormat` is a valid date until the component performs an update.

Do you think we could fail earlier? 
E.g. throw and exception during development if the `dateFormat` is incorrect.

src/lib/components/dateSelector/index.test.tsx
```diff
@@ -14,40 +14,138 @@ const setup = (date: string, onChange: (date: string) => void = () => {}) => {
...
+    expect(getByTestId('date-selector-year')).toHaveValue('2024');
+  });
+
+  it('should call onChange with the value changed', async () => {
+    const callback = jest.fn();
+    const { getByTestId, user } = setup(undefined, callback);
+
+    await user.type(getByTestId('date-selector-day'), '5');
+    await user.type(getByTestId('date-selector-month'), '7');
+    await user.type(getByTestId('date-selector-year'), '2023');
```
talo242, line 32: To make sure we are not impacting accessibility, IMHO it'd be better if we stick to the actual input labels, and not rely on `data-testid`.

Is there a reason to use `data-testid` to target these inputs?

src/lib/components/dateSelector/index.tsx
```diff
@@ -59,229 +24,233 @@ export interface DateSelectorProps {
...
+  }
 
-    setOpenCalendar(false);
-  };
+  const getInputProps = (key: keyof CalendarDate, index: number) => ({
+    'data-cy': `date-selector-${key}`,
+    'data-testid': `date-selector-${key}`,
+    className: styles[`${key}Input`],
+    id: key,
+    name: key,
```
talo242, line 199: How can we use this as uncontrolled component?

To give you more context, on the task engine we are adding a specific name and id to each input.
For example, a text input would have an id like this: `input-f8556724-a703-455e-a6f5-16b0d16bb26d`

But for the date input, we want the full date, so we are doing a tiny hack [by using a hidden field with the full date](https://github.com/getPopsure/app/blob/e2069cacaed963aaeb99f0afa3a122f0a7cf7b6d/src/features/taskEngine/components/CustomTask/inputs/DateInput.tsx#L17).

I'm wondering if we should have such hidden field already inside the component instead of the individual name for each input.

Happy to discuss this a bit further. We don't need to change anything for this PR.

src/lib/components/dateSelector/index.tsx
```diff
@@ -59,229 +24,233 @@ export interface DateSelectorProps {
...
+    maxLength: 2,
+    required: true,
+    label: placeholders?.[key],
+    placeholder: placeholders?.[`${key}Format` as FormatPlaceholder] ?? "",
+    labelInsideInput: true,
```
talo242, line 204: Are we not allowing this to be configured via props?

talo242 (changes requested):
Works pretty well!

I'm requesting changes mostly to use the input labels instead of the data-testid.

PR #315, opened 2024-05-16
refactor(multidropzone): [SIDE-7] Update MultiDropzone component to have default accept type

diogomateus:
### What this PR does
Updates `MultiDropzone` component to:
- have default accept type 
- have available video group types
- minor style fixes and message copy updates.

Solves:
SIDE-7

### How to test?

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/multiDropzone/index.test.tsx
+++ b/src/lib/components/multiDropzone/index.test.tsx
@@ -56,7 +56,7 @@ describe('MultiDropzone component', () => {
         ],
       });
 
-      expect(screen.getByText('Too many files.')).toBeVisible();
+      expect(screen.getByText('You can upload maximum 1 files.')).toBeVisible();
     });
 
     it('should show max file size error message', async () => {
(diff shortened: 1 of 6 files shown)


src/lib/components/multiDropzone/types.ts
```diff
@@ -1,25 +1,41 @@
 import { Accept } from "react-dropzone";
 
 export enum FileMimeTypes {
+  avi = "video/x-msvideo",
```
kimkwanka, line 4: 👴

kimkwanka (approved):

src/lib/components/multiDropzone/index.test.tsx
```diff
@@ -128,14 +128,14 @@ describe('MultiDropzone component', () => {
       const screen = setup({ accept: 'image' });
 
       expect(
-        screen.getByText('Supports HEIC, BMP, JPEG, JPG, PNG')
+        screen.getByText('Supports images, videos and documents')
```
talo242, line 131: Shouldn't this show only "images" because we are explicitly adding the `{ accept: 'image' }` object?

It seems like we are dropping the labelling based on the file types. Is that what we want?

talo242 (changes requested):

src/lib/components/multiDropzone/index.test.tsx
diogomateus: @talo242 Fixed it. For custom accept types you need to provide your own label but for generic image/video/documents it will now show the default label.

talo242 (approved):

PR #317, opened 2024-06-07
feat(CoverageTable): [EPIC-137] Improve CoverageTable

kimkwanka:
### What this PR does

This PR improves the CoverageTable implementation by doing the following:
* Add `hideScrollBarsMobile` prop
* Fix scrollbar hiding killing the mobile view buttons
* Add props to change detail button captions
* Truly disable `TableArrows` buttons to avoid some edge cases
* Expose `onSelectionChanged` callback
* Replace underline links with (i) buttons
* Expose `classNameOverrides` and made "sections" into real `section`s
* Increase cell text contrast
* Fixed an issue where uncollapsed sections would overflow on the right
* Adjust padding and cell default sizes
* Fix addon card size

### Why is this needed?

To have a more streamlined CoverageTable which still offers flexibility.

Solves:
EPIC-137

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Edge

### DIFF
--- a/src/lib/components/comparisonTable/components/Row/style.module.scss
+++ b/src/lib/components/comparisonTable/components/Row/style.module.scss
@@ -31,10 +31,10 @@ $cell-min-width: var(--cellWidth, 211px); // 195 + 16px
 
   line-height: 24px;
 
-  padding: 16px 0;
+  padding: 16px 8px;
 
   &:first-child {
-    padding: 16px 0 16px 16px;
+    padding: 16px 8px 16px 16px;
   }
 
   color: $ds-grey-900;
@@ -60,10 +60,10 @@ $cell-min-width: var(--cellWidth, 211px); // 195 + 16px
     flex: 1 0 $cell-min-width;
     width: $cell-min-width;
     &:first-child {
-      padding: 24px 0 24px 24px;
+      padding: 24px 8px 24px 24px;
     }
 
-    padding: 24px 0;
+    padding: 24px 8px;
   }
 }
 
--- a/src/lib/components/comparisonTable/components/Row/style.module.scss
+++ b/src/lib/components/comparisonTable/components/Row/style.module.scss
@@ -1,7 +1,7 @@
 @use "../../../../scss/public/colors" as *;
 @use "../../../../scss/public/grid" as *;
 
-$cell-min-width: var(--cellWidth, 256px);
+$cell-min-width: var(--cellWidth, 211px); // 195 + 16px
 
 .header {
   position: relative;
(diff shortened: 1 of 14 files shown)


src/lib/components/comparisonTable/components/Row/style.module.scss
```diff
@@ -116,7 +120,7 @@ h4.cell {
 h4.addon {
   border-right: none;
   max-width: calc(100vw - 64px);
-  width: 100vw;
+  width: 100&;
```
diogomateus, line 123: Is this a typo?

src/lib/components/comparisonTable/index.stories.mdx
```diff
@@ -27,17 +27,22 @@ The Comparison Table component provides an easy way to compare vast amounts of i
...
+| collapsibleSections   | boolean                                   | Make table groups with a label collapsible                          | false          | false    |
+| cellWidth             | number                                    | Width of a table content cell                                       | 211px          | false    |
+| firstColumnWidth      | number                                    | Width of the first column of the table                              | 212px          | false    |
+| stickyHeaderTopOffset | number                                    | Y-offset of the sticky header row                                   | 0px            | false    |
+| growContent           | boolean                                   | Makes the content area grow to fill the available space             | false          | false    |
+| onSelectionChanged    | (number) => void                          | Callback to be called when the selected tab index changes on mobile | n/a            | false    |
+| classNameOverrides    | [ClassNameOverrides](#classnameoverrides) | "className" props to be passed down to various internal components  | n/a            | false    |
```
diogomateus, line 45: Eventually would be cool to move this to JSX stories but we can wait for a major refactor.

src/lib/components/comparisonTable/components/Row/style.module.scss
kimkwanka: Yup, but if it works without I guess, then it's not even needed :D

src/lib/components/comparisonTable/index.stories.mdx
kimkwanka: Yeah true, didn't think about that!

src/lib/components/comparisonTable/components/Row/style.module.scss
kimkwanka: But it didn't work without it 😂

diogomateus:
My only question is about these two blocks having different widths, is this expected?

Screenshot 2024-06-07 at 15 14 11

diogomateus (approved):

kimkwanka:
> My only question is about these two blocks having different widths, is this expected?
> 
> Screenshot 2024-06-07 at 15 14 11

Yeah it's expected, as that details button just spans the whole container width whereas the sections are as big as their internal content. I'll try to figure out something. Thanks for bringing this up!

kimkwanka:
I exposed the details button's className as well now, so that one can add custom styling. @diogomateus

PR #322, opened 2024-06-17
refactor(card-button): Refactor CardButton and reorganize cards stories

diogomateus:
### What this PR does
This PR refactors `CardButton` to use `Card` component under the hood and moves Card component stories inside Stories folder.

### How to test?
Visit http://localhost:9009/?path=/story/jsx-cards-cardbutton--card-button-story

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge


kimkwanka:
While not 100% related, the new Card component's content is centered by default which makes no difference in isolation but when the card height increases (due to different length of cards in the same container for example), it becomes apparent.

In the designs I encountered it looks like the default behaviour should be "flex-start" instead however:

image


image

Do you think this is something that might make sense to add this default behaviour to the Card component? (Same with the icon on the left hand side which is also not aligned flex start)

kimkwanka (approved):

PR #323, opened 2024-06-18
feat(accordion): Add Accordion component

diogomateus:
### What this PR does
This PR adds the `Accordion` component and does minimal changes to `Card` component for improved rewriting of styles.

### How to test?
Run storybook book and visit [Accordion story page](http://localhost:9009/?path=/docs/jsx-accordion--accordion-story).

**Preview:**
https://github.com/getPopsure/dirty-swan/assets/4015038/7d5a210f-e6f4-41f0-93eb-a95f91b362bb

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/accordion/index.tsx
+++ b/src/lib/components/accordion/index.tsx
@@ -5,13 +5,12 @@ import AnimateHeight from 'react-animate-height';
 import styles from './style.module.scss';
 import { ChevronDownIcon } from '../icon';
 import { Card } from '../cards/card';
-import { Markdown } from '../markdown';
 
 export interface AccordionItem {
   id: string;
   question: string;
   icon?: ReactNode;
-  answer: string;
+  answer: ReactNode;
 }
 
 export interface AccordionProps {
@@ -63,7 +62,6 @@ const Accordion = ({
     <div
       className={
         classnamesUtil(
-          styles.wrapper,
           {
             'br8': !isDefaultVariant,
             [styles.wrapperBordered]: !isDefaultVariant,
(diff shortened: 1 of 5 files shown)


diogomateus:
Missing design feedback on hover/focus states but ready for review apart from that

src/lib/components/accordion/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+              height={isOpen ? 'auto' : 0}
+            >
+              <div 
+                className={classnamesUtil(
+                  classNames?.answer,
+                  'tc-grey-600 p16 pl0', 
+                  { 'pl16': !isDefaultVariant },
+                )}
+              >
+                {answer}
```
diogomateus, line 137: There is no Markdown integration, that needs to be done in other repositories as it differs between website/app.

src/lib/components/accordion/style.module.scss
```diff
@@ -0,0 +1,39 @@
...
+    border-bottom: 0;
+  }
+}
+
+.buttonWrapper {
+  border-radius: 0!important;
+
+  &:hover {
+    background: $ds-primary-50;
+    outline: 0 solid transparent!important;
```
diogomateus, line 24: Sadly we do need these `!important` here to override the current styles.

src/lib/components/accordion/index.tsx
kimkwanka: Can you elaborate?

diogomateus: Sure! On both [app](https://github.com/getPopsure/app/blob/main/src/components/markdown/index.tsx) and [website](https://github.com/getPopsure/website/blob/main/components/markdown/index.tsx) we have different custom markdown integrations so we'll need to configure components to have their own markdown on answers.

I'm talking about this part that wasn't implemented on design system in this PR:
https://github.com/getPopsure/app/blob/main/src/components/faq/index.tsx#L77

a-maffei (approved):

a-maffei:
> Missing design feedback on hover/focus states but ready for review apart from that

I was wondering about the focus state of the first item (whether only the bottom margin should have a border, unlike the other items), but it seems you'll tackle it separately. Other than that, it looks good! 💅

image 10

PR #356, opened 2024-07-16
feat(table): [SIDE-349] Add Table component

diogomateus:
### What this PR does

1.
2.
3.

### Why is this needed?

Please include relevant motivation and context of why this PR is necessary, sentry/linear/notion/...

Solves:
STO-XXX
EMU-XXX
RVN-XXX

### How to test?

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/table/components/TableCell/index.tsx
+++ b/src/lib/components/table/components/TableCell/index.tsx
@@ -1,0 +1,106 @@
+import { HTMLProps, ReactNode } from "react";
+import { Button } from "../../../button";
+import {
+  CheckIcon,
+  InfoIcon,
+  StarFilledIcon,
+  XIcon,
+  ZapFilledIcon
+} from "../../../icon";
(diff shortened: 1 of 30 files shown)


src/lib/components/table/components/TableCell/index.tsx
```diff
@@ -0,0 +1,106 @@
+import { HTMLProps, ReactNode } from "react";
```
kimkwanka, line 1: This goes for all of the components:
As discussed in one of our FEJFs we should not use `index.tsx` and `style.module.scss` and instead use `ComponentName.tsx` and `ComponentName.module.scss` to make it easier to reason what you're working on. The final export from DS can stay as is, but all other internal exports and imports don't need an index file.

kimkwanka (approved):

PR #385, opened 2024-08-19
fix(table): Remove pointer events none from ButtonCell

diogomateus:
### What this PR does
Remove pointer events none from ButtonCell as this prevents the button from being clicked on mobile.
### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/table/components/TableCell/ButtonCell/ButtonCell.module.scss
+++ b/src/lib/components/table/components/TableCell/ButtonCell/ButtonCell.module.scss
@@ -18,10 +18,6 @@
       background-color: $ds-grey-200;
     }
   }
-
-  @include p-size-mobile {
-    pointer-events: none;
-  }
 }
 
 .withoutPrice {


src/lib/components/table/components/TableCell/ButtonCell/ButtonCell.module.scss
```diff
@@ -18,10 +18,6 @@
       background-color: $ds-grey-200;
     }
   }
-
-  @include p-size-mobile {
-    pointer-events: none;
```
kimkwanka, line 23: This was intentionally added, can you explain why we need to remove this?

In the design's I have seen so far, ButtonCell was only ever used as a "selector button" for selecting a tariff or something. On desktop it's fine to let the button itself trigger this. On mobile however, the selection is triggered via the arrow keys, not the "button" in the center. Letting the button in the middle behave like a button in those cases lead to weird edge cases.

diogomateus: Oh, I see, I was trying to click the button but it wasn't working, I didn't realise I had to manually set this up using `onSelectionChanged`. I'll close this PR and update my code to use it instead.

PR #391, opened 2024-08-30
refactor(date-selector): [SIDE-380] Disable DateSelector auto-advance

diogomateus:
### What this PR does
This PR removes the auto-advance from DateSelector where a cursor would move to the next input if reaching end of current input (e.g. typed 12 on day would move cursor automatically to month or typing 4 in month would move cursor to year).

Solves:
SIDE-380

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/dateSelector/index.test.tsx
+++ b/src/lib/components/dateSelector/index.test.tsx
@@ -82,7 +82,7 @@ describe('DateSelector component', () => {
     expect(callback).not.toHaveBeenCalled();
   });
 
-  it('should show error boundaries error for ,ax date onChange empty when year out of boundaries', async () => {
+  it('should show error boundaries error for max date onChange empty when year out of boundaries', async () => {
     const callback = jest.fn();
     const date = '2100-01-01';
     const { getByTestId } = setup(date, callback);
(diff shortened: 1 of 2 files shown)


talo242 (approved):
Tested locally and works as expected.
Can I ask why did we decide to move away from this behavior?

diogomateus:
> Tested locally and works as expected. Can I ask why did we decide to move away from this behavior?

I'll have to update this as it seems this is only intended to be disabled on non mobile inputs 🤔 you can read more about the rationale behind disabling this [here](https://linear.app/feather-insurance/issue/SIDE-380/disable-auto-advance)

busypeoples:
> > Tested locally and works as expected. Can I ask why did we decide to move away from this behavior?
> 
> I'll have to update this as it seems this is only intended to be disabled on non mobile inputs 🤔 you can read more about the rationale behind disabling this [here](https://linear.app/feather-insurance/issue/SIDE-380/disable-auto-advance)

Does this work for mobile or does this PR have to be updated to keep the auto-advance for mobile?

PR #394, opened 2024-09-20
fix(table): [SIDE-637] do not push info button to next column

talo242:
### What this PR does

This PR updates the `<Table />` component to prevent long strings of text to push info buttons:

| Before | After |
|--------|--------|
| Screenshot 2024-09-20 at 10 14 20 | Screenshot 2024-09-20 at 10 13 18 | 

### Why is this needed?

To keep info buttons on the same column as the content

Solves:
SIDE-637

### How to test?
You can check the new behavior [on Storybook](http://localhost:9009/?path=/story/jsx-table--table-story&globals=viewport:mobile1). You can try and change the text length of the cells content and see how the info button behaves on both mobile and desktop.

### Checklist:

- [x] I reviewed my own code
- [ ] 

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Edge

### DIFF
--- a/src/lib/components/table/components/TableCell/BaseCell/BaseCell.tsx
+++ b/src/lib/components/table/components/TableCell/BaseCell/BaseCell.tsx
@@ -1,8 +1,8 @@
 import classNames from 'classnames';
 import {
   CheckIcon,
-  XIcon,
   StarFilledIcon,
+  XIcon,
   ZapFilledIcon,
 } from '../../../../icon';
 import { ReactNode } from 'react';
@@ -80,7 +80,8 @@ export const BaseCell = ({
         className={classNames(
           'd-flex fd-column',
           alignClassName,
-          styles.relative
+          styles.relative,
+          { [styles.maxWidth]: align === 'center' }
         )}
       >
         {progressBarValue !== undefined && (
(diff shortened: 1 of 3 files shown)


src/lib/components/table/components/TableCell/BaseCell/BaseCell.tsx
```diff
@@ -80,7 +80,8 @@ export const BaseCell = ({
         className={classNames(
           'd-flex fd-column',
           alignClassName,
-          styles.relative
+          styles.relative,
+          { [styles.maxWidth]: align === 'center' }
```
diogomateus, line 84: Do you think this should only be applied when there is modalContent?

talo242: What you mean is not applying this style if the content is a Checkmark or Rating?

diogomateus: Yes, I thin Checkmark or Rating should be used with a description that's shown below, not on the side, what do you think?

talo242: Ok I get what you mean. I'll make the change.

diogomateus (approved):
Thank you for looking into this 🙌

PR #396, opened 2024-09-24
feat(button): [SIDE-647] introduce the `filledBlack` variant

a-maffei:
### What this PR does

This PR introduces a new button variant (`filledBlack`) -> [Figma](https://www.figma.com/design/MKs4cbojdVOBKUxv7okb93/Dirty-Swan-Design-System?node-id=17167-87383&t=O1Hc6mhD7LNMUBMU-4)

image 16

image 15

### Why is this needed?

The Design team is introducing this variant in our system. 
We need to implement this variant with high priority in the B2B website banner (as a start). 

Solves:
SIDE-647

### How to test?

You can check the new button variant locally [here](http://localhost:9009/?path=/story/jsx-button--button-variants), also for the hover, focus, and disabled state. 

### Checklist:

- [x] I reviewed my own code

### Browser support

- [x] Firefox
- [x] Chrome
- [x] Safari
- [x] Edge

### DIFF
--- a/src/lib/components/button/index.stories.tsx
+++ b/src/lib/components/button/index.stories.tsx
@@ -76,7 +76,7 @@ export const ButtonVariants = ({ children, onClick }: ButtonProps) => (
   <div>
     <h3 className='p-h3 mb24'>Filled variants</h3>
     <div className='d-flex gap16 p24 bg-grey-300 br8'>
-      {[ "filledColor", "filledGray", "filledWhite"].map((variant) => (
+      {[ "filledColor", "filledGray", "filledWhite", "filledBlack"].map((variant) => (
           <div key={variant}>
             <h4 className='p-h4 mb16'>
               {variant}
--- a/src/lib/components/button/index.tsx
+++ b/src/lib/components/button/index.tsx
@@ -5,6 +5,7 @@ type ButtonVariant =
   | 'filledColor'
   | 'filledGray'
   | 'filledWhite'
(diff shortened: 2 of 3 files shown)


diogomateus (changes requested):
Just two minor things:
- disabled state background color should stay the same on hover
- focus state isn't very accessible (border/outline is black), should we use the same primary color or something else?

a-maffei:
> Just two minor things:
> 
> * disabled state background color should stay the same on hover
> * focus state isn't very accessible (border/outline is black), should we use the same primary color or something else?

1) Makes sense, will update!
2) I got this focus state from [Figma](https://www.figma.com/design/MKs4cbojdVOBKUxv7okb93/Dirty-Swan-Design-System?node-id=251-26&node-type=frame&t=O1Hc6mhD7LNMUBMU-0). We can use the primary color and let Andrey know.

diogomateus:
Disabled is now having a color primary that's not visible:
Screenshot 2024-09-24 at 11 59 17
and hover state is white
Screenshot 2024-09-24 at 11 59 20

a-maffei:
> Disabled is now having a color primary that's not visible: Screenshot 2024-09-24 at 11 59 17 and hover state is white Screenshot 2024-09-24 at 11 59 20

I wasn't ovverriding the secondary button style correctly. It should be all fixed now - thank you for spotting this!

diogomateus (approved):
🏴 🏴 🏴

PR #397, opened 2024-10-01
refactor(table): update info button positioning

talo242:
### What this PR does

This PR updates the position of table's info button component, so it doesn't get pushed whenever the table cell has long content.

**Before**
372394235-decd6b23-4f1c-485b-ab19-c278a907ba06

**After**
Screenshot 2024-10-01 at 10 58 22

### DIFF
--- a/src/lib/components/table/Table.stories.tsx
+++ b/src/lib/components/table/Table.stories.tsx
@@ -162,7 +162,11 @@ const initialData: TableData = [
       ],
       [
         { text: 'Operations', modalContent: 'info' },
-        { checkmarkValue: true, modalContent: 'Maybe' },
+        {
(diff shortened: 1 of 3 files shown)


diogomateus (approved):
Tested it out and works very well, thank you for improving this 🙌 

Quick question, is this the expected position for the text only when breaking lines?
Screenshot 2024-10-01 at 10 31 18

talo242:
> Tested it out and works very well, thank you for improving this 🙌
> 
> Quick question, is this the expected position for the text only when breaking lines? Screenshot 2024-10-01 at 10 31 18

@diogomateus thanks for the review!
I think that's a better positioning, it behaves more similar to the table header and also it sticks closer to the text.

PR #402, opened 2024-11-25
refactor(a11y): Update accessibility in components

diogomateus:
#### AutoComplete address
- Add a way to go back to search after manual entry
#### Badge
- Add role for better accessibillity
#### Segmented Control
- Improve focus styles and allow navigating with space key
#### Spinner
- Add aria properties

### Checklist:

- [ ] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/autocompleteAddress/index.tsx
+++ b/src/lib/components/autocompleteAddress/index.tsx
@@ -80,6 +80,7 @@ export interface AutocompleteAddressProps {
   manualAddressEntryTexts?: {
     preText?: string;
     cta?: string;
+    ctaSearch?: string;
   };
   countryCode?: string;
 }
@@ -250,7 +251,7 @@ const AutocompleteAddress = ({
           [styles['map-container--hidden']]: place === null,
         })}
       >
-        <div className={styles.map} id="map" />
+        <div className={styles.map} id="map" aria-hidden="true" />
         {isLoading && (
           <div className={styles['loading-spinner']}>
             <div className="ds-spinner ds-spinner__m" />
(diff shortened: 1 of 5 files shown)


src/lib/components/autocompleteAddress/index.tsx
```diff
@@ -250,15 +251,15 @@ const AutocompleteAddress = ({
           [styles['map-container--hidden']]: place === null,
         })}
       >
-        <div className={styles.map} id="map" />
+        <div className={styles.map} id="map" aria-hidden="true" />
```
a-maffei, line 254: This is also a more general question: if we're hiding the loading status visual cue to assistive technologies, shall we find another way to communicate it ([like in this example](https://carbondesignsystem.com/components/loading/accessibility/#development-considerations) or in some of the changes you made in this PR)?

src/lib/components/spinner/index.tsx
```diff
@@ -6,6 +6,7 @@ export interface SpinnerProps {
 
 const Spinner = ({ size = 's' }: SpinnerProps) => (
   <div
+    aria-live='polite'
```
a-maffei, line 9: Do I understand correctly, that aria-live needs [some text or message that changes dynamically and should be announced](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live#description)? If so, what would it be in this case?

a-maffei (approved):

src/lib/components/autocompleteAddress/index.tsx
diogomateus: This is just the map preview which is just visual and doesn't provide much value for screen readers

PR #413, opened 2025-01-24
Revert "fix(table): [SIDE-1316] Fix first row hidden on scrolling to  anchor"

diogomateus:
Reverts getPopsure/dirty-swan#412


diogomateus:
@talo242 Reverting as this PR is causing unwanted additional space on top of  comparison table.

PR #422, opened 2025-05-26
feat(a11y): [SIDE-1442] group radio/checkbox buttons

a-maffei:
### What this PR does

1. Wraps radio and checkbox inputs in a `<fieldset>` to group related controls
2. Adds a visually hidden `<legend>` for screen reader context
3. Adds a `name` attribute to radio buttons belonging to the same group

### Why is this needed?

Improves accessibility by allowing screen readers to treat radio and checkbox options as a one labeled group.

radio

radio-html

Solves:
SIDE-1442

### How to test?

- Inspect the HTML of the corresponding Storybook components
- Use the components with a screen reader: listen for a group label being read and the number of options in the group.

### Checklist:

- [x] I reviewed my own code

### Browser support

- [ ] Firefox
- [x] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/input/checkbox/index.tsx
+++ b/src/lib/components/input/checkbox/index.tsx
@@ -20,6 +20,7 @@ export interface CheckboxProps<ValueType extends string> {
     label?: string;
     option?: string;
   };
+  fieldLegend?: string;
 }
 
 export const Checkbox = <ValueType extends string>({
@@ -30,6 +31,7 @@ export const Checkbox = <ValueType extends string>({
   inlineLayout = false,
   bordered = true,
   classNames: classNamesObj,
+  fieldLegend,
 }: CheckboxProps<ValueType> & {}) => {
   const hasNoneValue = Object.keys(options).includes('NONE');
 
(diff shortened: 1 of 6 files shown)


src/lib/components/input/checkbox/styles.module.scss
```diff
@@ -1,5 +1,9 @@
 .container {
   max-width: 100%;
+  border: 0;
```
a-maffei, line 3: Source for the `<fieldset>` reset: https://thatemil.com/blog/2015/01/03/reset-your-fieldset/

src/lib/components/input/checkbox/index.tsx
```diff
@@ -30,6 +31,7 @@ export const Checkbox = <ValueType extends string>({
   inlineLayout = false,
   bordered = true,
   classNames: classNamesObj,
+  fieldLegend,
```
diogomateus, line 34: Doesn't really make a difference but you can also set the default here:
fieldLegend = 'Select one or more options',

diogomateus (approved):
Left a minor comment but it's optional, looks great! 🙌

kimkwanka (approved):

PR #434, opened 2025-08-08
refactor(card): [EPIC-709] disable hover and support active, focus statuses

sk8Guerra:
### What this PR does

**Style and Accessibility Updates:**

* The card's hover and focus outlines have been updated: hover now uses a grey outline instead of primary color, and focus uses a darker grey outline for better accessibility contrast.
* Additional styles were added for coarse pointer devices (like touchscreens): hover outlines are removed, and active/focus states use the new grey outline and primary text color, improving the touch experience.

### FIgma

[link](https://www.figma.com/design/MKs4cbojdVOBKUxv7okb93/Dirty-Swan-Design-System?node-id=18661-99812&t=4d7UkDaiwvcKOoyA-0)

Screenshot 2025-08-08 at 11 26 49

### Why is this needed?

Mainly to disable hover state in mobile and support active and focus state.

Solves:
EPIC-709

### Checklist:

- [x] I reviewed my own code

### Browser support

- [ ] Firefox
- [ ] Chrome
- [ ] Safari
- [ ] Edge

### DIFF
--- a/src/lib/components/cards/card/style.module.scss
+++ b/src/lib/components/cards/card/style.module.scss
@@ -9,13 +9,27 @@
   border-radius: 8px;
 
   &:hover {
-    outline: 1px solid $ds-primary-500;
+    outline: 1px solid $ds-grey-500;
     color: $ds-primary-500;
   }
 
+  &:active,
+  &:focus,
   &:focus-visible {
-    outline: 2px solid $ds-primary-500;
+    outline: 2px solid $ds-grey-900;
     color: $ds-primary-500;
+  }
+
+  @media (pointer: coarse) {
+    &:hover {
+      outline: none;
+    }
+
+    &:active,
+    &:focus {
+      outline: 2px solid $ds-grey-900;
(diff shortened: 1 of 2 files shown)


src/lib/components/cards/card/style.module.scss
```diff
@@ -9,14 +9,28 @@
   border-radius: 8px;
 
   &:hover {
-    outline: 1px solid $ds-primary-500;
+    outline: 1px solid $ds-grey-500;
```
theerebuss, line 12: If we don't have one yet, IMO it would make sense to create a hover and focus token that we reuse here, instead of using a specific color. Whenever we change the brand, it would be harder to spot these edge cases.

diogomateus: Agreed, I'll work on color updates soon and we might implement that, I'll sync with design on it!

sk8Guerra:
I will leave this PR without merge for now. I will take this work then working in the redesign in a future sprint.

PR #508, opened 2026-06-01
refactor: set latest tag when publishing npm package

talo242:
### What this PR does

1. Explicitly adds the `latest` tag when deploying the package
2. Sets the `package-manager-cache` to false as suggested by the [NPM documentation](https://docs.npmjs.com/generating-provenance-statements#example-github-actions-workflow)

### Why is this needed?

We have some stray higher versions of the package (27.00.xx) that were published by mistake.
NPM does not allow implicitly setting up a lower version as `latest` so we have explicitly pass this option.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -33,12 +33,13 @@ jobs:
           node-version-file: '.nvmrc'
           registry-url: 'https://registry.npmjs.org'
           scope: '@popsure'
+          package-manager-cache: false # never use caching in release builds
       - name: Set package version
         run: npm version $RELEASE_VERSION --no-git-tag-version --allow-same-version
       - name: Publish new version
         # No NODE_AUTH_TOKEN: npm authenticates via the OIDC token exchange.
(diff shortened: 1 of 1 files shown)


eneax (approved):

theerebuss:
Can we not delete the new versions? Wondering if shipping with `latest` won't bump the version to > 27.x.x

theerebuss (approved):

talo242:
> @theerebuss Can we not delete the new versions?

I think we can, but we need to unpublish+delete them. It will be nice to do that clean up soon, but for now it's out of scope.
For reference, we need to remove this versions: 27.0.0, 27.0.1, 27.0.2, 27.0.5

> Wondering if shipping with latest won't bump the version to > 27.x.x 

Edit: I checked and it should add the correct version. This is the error that this PR is trying to fix:
```
npm error Cannot implicitly apply the "latest" tag because previously published version 27.0.1 is higher than the new version 0.66.9. You must specify a tag using --tag.
```

PR #511, opened 2026-07-22
refactor(table): [PRO-7028] Show selected column

diogomateus:
### What this PR does
Show selected column based on the combination of `activeSection` and `showSelectedColumn` props.

Screenshot 2026-07-22 at 11 38 03

https://github.com/user-attachments/assets/b26c975e-7af6-48c2-8e16-1b014dd0bbf9

### DIFF
--- a/src/lib/components/table/components/TableCell/TableCell.tsx
+++ b/src/lib/components/table/components/TableCell/TableCell.tsx
@@ -13,6 +13,8 @@ type ExtraTableCellProps = {
   isFirstCellInRow?: boolean;
   isTopLeftCell?: boolean;
   isNavigation?: boolean;
+  isSelectedColumn?: boolean;
+  selectedColumnPosition?: 'top' | 'bottom' | 'middle';
   imageComponent?: (args: any) => JSX.Element;
   isBelowDesktop?: boolean;
 };
@@ -24,6 +26,8 @@ const TableCell = React.memo(
     isFirstCellInRow = false,
     isHeader = false,
     isNavigation = false,
+    isSelectedColumn = false,
+    selectedColumnPosition,
     isTopLeftCell = false,
     colSpan = 0,
     isBelowDesktop,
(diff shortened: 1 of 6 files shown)


src/lib/components/table/components/TableSection/TableSection.tsx
```diff
@@ -36,10 +37,21 @@ const TableSection = ({
...
+  const firstVisibleBodyRowIndex = tableCellRows.findIndex(
+    (_, index) => index > 0 && !hideRows.includes(index)
+  );
+
+  const lastVisibleRowIndex = tableCellRows.reduce(
+    (last, row, index) =>
+      index > 0 && !hideRows.includes(index) && row.length > 1 ? index : last,
+    -1
+  );
+
```
sk8Guerra, line 54: What is the purpose of this?

src/lib/components/table/components/TableCell/TableCell.tsx
```diff
@@ -13,6 +13,8 @@ type ExtraTableCellProps = {
   isFirstCellInRow?: boolean;
   isTopLeftCell?: boolean;
   isNavigation?: boolean;
+  isSelectedColumn?: boolean;
+  selectedColumnPosition?: 'top' | 'bottom' | 'middle';
```
sk8Guerra, line 17: Why do we need a position here? and these are the two new special props we would need to support in the cms right?

src/lib/components/table/components/TableSection/TableSection.tsx
diogomateus: We need to know the first and last rows of each section so we can apply the top and bottom border radius 🙃

src/lib/components/table/components/TableCell/TableCell.tsx
diogomateus: This is internal only and just to show the bottom and bottom border radius i mentioned on the comment above

sk8Guerra (approved):

PR #513, opened 2026-07-24
fix(table): [PRO-7028] Remove vertical divider and left align action cells

sk8Guerra:
### What this PR does

Two follow-up adjustments to the Table redesign shipped in 0.66.13, based on design review of the multi-plan table:

- Removes the vertical divider line that separated the sticky label column from the plan columns. It no longer exists in the new design.
- Left-aligns the action cells, matching the left alignment of all other cell content: the CTA cells (plan title, price, and call-to-action button) and the plan selection button cells were previously centered in their columns. The buttons keep their existing width constraints, anchored left.

Screenshot 2026-07-24 at 11 05 33


diogomateus (approved):
🥳

PR #520, opened 2026-08-04
feat(tabs): Add Tabs component and use it on mobile Table header

diogomateus:
### What this PR does
  1. Adds a new Tabs component (src/lib/components/tabs/) with full keyboard navigation and ARIA tablist support
  2. Integrates Tabs into the Table component as an alternative mobile navigation mode via a new mobileNavigationMode prop ('arrows' | 'tabs')
  3. Exports Tabs, TabsProps, and TabItem from the library index

###  Why is this needed?
  On mobile, the Table component currently uses left/right arrow buttons to navigate between columns. This adds a tab-based navigation alternative that gives users a clearer overview of all available columns at a glance, improving discoverability and usability on mobile devices.

###  How to test?
  - Open the Tabs stories in Storybook (JSX/Tabs) to verify the standalone component (Default, Two Tabs, Many Tabs)
  - Open the Table > Tabs Navigation (Mobile) story, resize the browser to a mobile width, and verify:
    - Tabs appear instead of arrow navigation
    - Clicking a tab switches the visible column
    - Keyboard navigation (ArrowLeft/ArrowRight) works between tabs
    - The tabs container sticks to the top when scrolling
  - Verify that existing Table stories still use arrow navigation by default (no regression)

### Preview
Screenshot 2026-08-04 at 14 56 12
Screenshot 2026-08-04 at 14 55 36

### DIFF
--- a/src/lib/components/table/Table.stories.tsx
+++ b/src/lib/components/table/Table.stories.tsx
@@ -939,6 +939,25 @@ export const ThreePlanQuoteWebsite = {
   name: 'Three Plan - Website',
 };
 
+export const TabsNavigationMobile = {
+  render: () => (
+    <div style={{ maxWidth: 900 }}>
+      <p className="p-p tc-neutral-500 mb16">
+        Resize the browser to a mobile width to see tabs instead of
+        left/right arrow buttons.
+      </p>
+      <Table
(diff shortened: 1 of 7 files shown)


sk8Guerra (approved):
Looking great! ❤️

PR #525, opened 2026-08-28
feat(modal): Add DirtySwanProvider context to track modal open state

diogomateus:
### What this PR does
  1. Adds a DirtySwanProvider React context that tracks modal open/close state using a ref counter, firing onModalChange(true) when the first modal opens and onModalChange(false) when the last one closes.
  2. Adds a useModalTrack hook that automatically registers/unregisters modal visibility with the provider, and integrates it into GenericModal.
  3. Exports DirtySwanProvider, useDirtySwan, and useModalTrack from the library, with unit tests and Storybook stories.

### Why is this needed?
  Consumers of dirty-swan need a way to know when any modal is open (e.g. to disable background scroll, pause animations, or toggle app-level UI). There was no centralized mechanism for this — each consumer had to track modal state independently.

### How to test?
  - Wrap your app in <DirtySwanProvider onModalChange={cb}> and open/close modals — verify the callback fires correctly.
  - Open nested modals and confirm onModalChange(false) only fires when the last modal closes.
  - Check the Storybook stories under JSX/DirtySwanProvider for single and nested modal demos.

https://github.com/user-attachments/assets/aaad7809-ab90-401a-a29d-4fe7f4e38796


src/lib/context.tsx
```diff
@@ -0,0 +1,60 @@
...
+  onModalChange?: (isModalOpen: boolean) => void;
+}
+
+export const DirtySwanProvider = ({
+  children,
+  onModalChange,
+}: DirtySwanProviderProps) => {
+  const countRef = useRef(0);
+  const onModalChangeRef = useRef(onModalChange);
+  onModalChangeRef.current = onModalChange;
```
sk8Guerra, line 21: I just do not understand so much this part of the code, what is the idea behind this lines of code?

sk8Guerra (approved):
Looking great! I just left a question.

src/lib/context.tsx
diogomateus: refs prevent react from re-rendering compared to useState/useCallback which would cause re-renderings. Since this will be a top component on all apps that use the design system, the stability and avoiding unncessary re-renderings is the top priority

